PR #44, opened 2020-05-03
Replace Willow w/ swift-log, Bitrise w/ GH Actions

MaxDesiatov:
Willow (even its latest version) currently breaks the build with this error:

```
error: target at 'WAKit/.build/checkouts/Willow' contains mixed language source files; 
feature not supported
```

Apple's [`swift-log`](https://github.com/apple/swift-log.git) seems to be a well-supported alternative, it was quite easy to replace Willow with it.

GitHub Actions now test on both macOS and Linux.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,8 +12,6 @@ jobs:
         run: sudo xcode-select -switch /Applications/Xcode_11.4.1.app
       - name: Test in Debug
         run: swift test -c debug
-      - name: Test in Release
-        run: swift test -c release
 
   build-linux:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,24 @@
+name: Build and test
+
+on: [push, pull_request]
+
+jobs:
+  build-macos:
+    runs-on: macos-latest
+    steps:
+      # Checks-out the repo. More at: https://github.com/actions/checkout
+      - uses: actions/checkout@v2
+      - name: Select Xcode version
+        run: sudo xcode-select -switch /Applications/Xcode_11.4.1.app
+      - name: Test in Debug
+        run: swift test -c debug
+      - name: Test in Release
+        run: swift test -c release
+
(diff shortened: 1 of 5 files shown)


MaxDesiatov:
@akkyie now that GitHub Actions pass, could you remove the Bitrise webhook? I don't have permissions to disable that I'm afraid.

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Test in Debug
+        run: swift test -c debug
+
+  build-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Didstopia/SwiftAction@v1.0.2
+        with:
+          swift-action: test --enable-test-discovery
```
akkyie, line 26: I don't have much experience with it, but ubuntu-latest looks to have pre-installed Swift.
https://github.com/actions/virtual-environments/blame/master/images/linux/Ubuntu1804-README.md#L40
Do we still need to use Docker-ed one, to avoid any problem with it? Maybe to fix the version?

MaxDesiatov: Yeah, the benefit of the Docker image is that it specifies the exact OS version and Swift verions, while with the `ubuntu-latest` image we're at the mercy of GitHub Actions maintainers, who can update those versions without (or with little) warning. We know that `Didstopia/SwiftAction@v1.0.2` ships with Swift 5.2, `Didstopia/SwiftAction@v1.0.1` with Swift 5.1, we can even test the library with both versions, which I do in some of my own repositories to make sure that backward and forward compatibility is maintained.

MaxDesiatov:
Thanks, I've posted the answer as a review comment. By the way, do you prefer linear history with "Squash and merge" (which I see is disabled in this repository) or non-linear history with merge commits?

akkyie:
I'd rather want to keep commits as-is, but have no strong opinion about it. What would you say

MaxDesiatov:
No problem at all, I'll continue using merge commits for consistency 🙂

PR #66, opened 2023-11-10
Expose a way to allocate a memory from host

MaxDesiatov:
When using WasmKit, in certain cases we need `HostModule` to provide a host memory instance, as opposed to an already supported case, where a parsed module has its own guest memory.


kateinoigakukun (approved):
Thanks!

PR #67, opened 2023-11-13
Extract `ExecutionState` from the `Runtime` type

MaxDesiatov:
`Runtime` no longer manages the execution state of a guest program. Separating the execution state makes it easier to have a fresh stack for each invocation of an exported function. Also it's now safe to provide a way to access `Runtime` during `HostFunction` execution since no mutable execution state is exposed.

Additionally, tests were added to verify the new API surface.


MaxDesiatov:
Marking as a draft until CI issues are resolved.

kateinoigakukun (approved):
Thanks!

PR #68, opened 2023-11-14
Mark `Function` as `internal` and rename it as `GuestFunction`

MaxDesiatov:
This resolves a naming collision between the existing `Function` type declared in `Types/Module.swift` and `Runtime/Function.swift` that in a subsequent change fails the build with these errors:

```
WasmKit/Types/Module.swift:21:21: error: 'Function' is ambiguous for type lookup in this context
        functions: [Function] = [],
                    ^~~~~~~~
WasmKit/Execution/Runtime/Function.swift:1:15: note: found this candidate
public struct Function: Equatable {
              ^
WasmKit/Types/Module.swift:70:15: note: found this candidate
public struct Function {
```


kateinoigakukun (approved):
Thanks!

PR #69, opened 2023-11-15
Support reentrant function call between host/guest

MaxDesiatov:
Since `Runtime` manages the single execution state, it wrongly executes upper frame instructions if the call stack is `guest -> host -> guest`. This change creates a fresh execution state for each exported guest function invocation and adds test coverage to verify that it fixes the issue.


kateinoigakukun (approved):
Thanks!

PR #92, opened 2024-05-01
Drop Foundation dependency

kateinoigakukun:


Sources/WASI/WASI.swift
```diff
@@ -1,4 +1,12 @@
-import Foundation
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import Glibc
+#elseif os(Windows)
+import ucrt
+#else
+#error("Unsupported Platform")
+#endif
```
kabiroberai, line 9: How come we need to use libc here? Seems like the only use site is `open(..., O_DIRECTORY)` — can we not use `SystemPackage.FileDescriptor.OpenOptions.directory` instead?

kateinoigakukun: I wanted to keep using `memchr` because llvm recognizes the lib call and optimizes it.

kabiroberai: I see, what do you think about making a `memchr` shim in SystemExtras? Might be nice to isolate the C library stuff to that one spot. We can do the same for `O_DIRECTORY`, which seems like it might need to be re-declared by us on Windows (other platforms define `.directory` in SystemPackage).

kateinoigakukun: Yes, importing it through a C shim could be an option. Actually `memchr` is not a syscall and `SystemExtras` will be upstreamed to swift-system later, so `SystemExtras` might not be an appropriate place.
Once we upgrade the minimum supported Swift version to 6.0, we can import it via `@_extern(c)` without C module, but it's not feasible right now.

kabiroberai: Ah interesting, I would've assumed that `UnsafeRawBufferPointer.firstIndex(of:)` could be optimized to `memchr` under the hood but it looks like that isn't being done yet. It's being tracked via https://github.com/apple/swift/issues/63200.

kateinoigakukun: Yeah, it's unfortunate situation now 🥲

PR #153, opened 2024-10-23
WasmParser: Specialize more hot paths

kateinoigakukun:
With this change, we can achieve 17% speedup on stringify macro
benchmark. The performance number is same as the case with CMO but with
smaller binary size.

|            | baseline | CMO   | this change |
|------------|----------|-------|-------------|
| speed (ms) | 198.1    | 172.8 | 169.9       |
| size  (kb) [^1] | 1739     | 1923  | 1797        |

[^1]: swift-6.0.1-RELEASE, macOS 14.4.1 arm64

### DIFF
--- a/Sources/WasmParser/LEB.swift
+++ b/Sources/WasmParser/LEB.swift
@@ -1,10 +1,11 @@
+@usableFromInline
 enum LEBError: Swift.Error, Equatable {
     case overflow
     case integerRepresentationTooLong
     case insufficientBytes
 }
 
-@inline(__always)
+@inlinable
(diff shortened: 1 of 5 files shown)


kateinoigakukun:
No speed up but it just adds code size

|            | baseline | CMO   | this change | this change + CMO |
|------------|----------|-------|-------------|------------|
| speed (ms) | 198.1    | 172.8 | 169.9       | 170.3 |
| size  (kb) [^1] | 1739     | 1923  | 1797        | 1897 |

kkebo:
Thanks!

PR #182, opened 2025-06-06
Android support

marcprux:
This PR adds support for building and testing on Android. It mostly just adds `#if os(Android)` checks and some Android-specific fixes for things like `errno` and `O_SYNC`, which are implemented differently in Bionic than Linux/Musl.

Android does require swift-system 1.5+, so I made the dependency `upToNextMajor` from "1.3.0" rather than `upToNextMinor`. If that presents an issue, we could probably finagle some mechanism for having different swift-system version requirements based on the platform.

Test cases are run with [swift-android-action](https://github.com/marketplace/actions/swift-android-action), which installs the Android SDK and runs the test cases on an Android emulator. Some of the tests that expect hard-coded file paths had to be disabled because the Android emulator doesn't have access to the local filesystem. This could potentially be resolved by [copying](https://github.com/marketplace/actions/swift-android-action#test-resources-and-environment-variables) the necessary files, but would require re-working all the `#filePath` references to use either relative paths, or accept some sort of override file root (perhaps based on an environment variable), and seemed like it would be a more intrusive change than would be warranted.


kateinoigakukun (approved):
Thanks!

PR #186, opened 2025-06-13
Support exploring ".wat" files

pcbeard:
Convert parseWasm() to be a global function instead of an an instance method of the Run command. This wraps WasmKit.parseWasm() and supports both ".wat" and ".wasm" files.


kateinoigakukun (approved):
Thanks!

PR #187, opened 2025-06-18
Fix `poll_oneoff` and more I/O syscalls in WASI 0.1

MaxDesiatov:
Testing Swift Concurrency for WebAssembly requires presence of `poll_oneoff`, which is absent in WasmKit. This change provides an implementation for POSIX hosts according to the ABI specified in WebAssembly System Interface standard https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#poll_oneoff

rdar://149935761


MaxDesiatov:
> Did we bump up the minimum supposed macOS version in the Swift OSS toolchain?

For macOS it's now the same as in SwiftPM https://github.com/swiftlang/swift-package-manager/blob/d8bc782bbb84012f3cdff97a4a2afed9029cda25/Package.swift#L133

kateinoigakukun (changes requested):
Please update wasi-testsuite and ensure that https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/poll_oneoff_stdio.rs integration test passes. I think we need to write back events buffer and also check error after `poll` library call.

kateinoigakukun (approved):

Sources/WASI/Poll.swift
```diff
@@ -0,0 +1,89 @@
...
+#if canImport(Darwin)
+    import Darwin
+#elseif canImport(Glibc)
+    import Glibc
+#elseif canImport(Musl)
+    import Musl
+#elseif canImport(Android)
+    import Android
+#elseif os(Windows)
+    import ucrt
```
al45tair, line 13: Seems odd; I'm pretty sure Windows doesn't have `poll()`, or indeed real fds (the C library does emulate them to a degree, but `HANDLE`s are the Windows equivalent, and the equivalent of `poll()` is either `WaitForMultipleObjects` or I/O completion ports). Is this supposed to work for Windows?

MaxDesiatov: It has `throw WASIAbi.Errno.ENOTSUP` on Windows in the implementation a few lines below.

PR #198, opened 2025-10-07
Add `breakpoint` VM instruction

MaxDesiatov:
Per a discussion, following options are available for enabling breakpoints in WasmKit:

> One of the main difficulties in implementing breakpoint in WasmKit is that our internal instruction set is not a 1:1 map of Wasm instruction set, unlike IPInt in JSC. Therefore, we need a way to translate address between Wasm byte offset <-> internal ISeq.
>
>Option 1: Record address translation table during compiling wasm bytecode to internal ISeq when a first breakpoint is set to the function, then patch the internal ISeq to set breakpoint using the table.
We only need one-time re-compilation to record the table, so later breakpoint settings will be trivial, but it adds non-trivial memory footprint to hold the table.
>
>Option 2: Patch the original wasm bytecode to replace the target instruction with a special reserved breakpoint instruction and compile to the internal ISeq.
We don't need to hold the addr translation table but need to re-compile every time a new breakpoint is set.
>
> Option 3: Implement in-place interpreter along side with the current fast interpreter.
This is the simplest approach and what wamr and JSC are doing for their classic interpreter and IPInt respectively, but need to maintain two interpreters.

We've decided to proceed with Option 2, where a special `breakpoint` instruction is added to the VM, which is translated from 0xFF Wasm binary code. This code can never be supplied by the user, as it isn't enabled in the parser. In a future PR we'll provide `public` methods on module instances like `func enableBreakpoint` that flips this instruction in in-memory module source and recompiles the corresponding function to VM bytecode with the new `breakpoint` instruction.

### DIFF
--- a/Sources/WasmKit/Translator.swift
+++ b/Sources/WasmKit/Translator.swift
@@ -820,7 +820,9 @@ struct InstructionTranslator: InstructionVisitor {
     /// The index of the function in the module
     let functionIndex: FunctionIndex
     /// Whether a call to this function should be intercepted
-    let intercepting: Bool
+    let isIntercepting: Bool
+    /// Whether Wasm debugging facilities are currently enabled.
+    let isDebugging: Bool
     var constantSlots: ConstSlots
     let validator: InstructionValidator
 
@@ -833,7 +835,8 @@ struct InstructionTranslator: InstructionVisitor {
         locals: [WasmTypes.ValueType],
         functionIndex: FunctionIndex,
         codeSize: Int,
-        intercepting: Bool
+        isIntercepting: Bool,
+        isDebugging: Bool = false
     ) throws {
         self.allocator = allocator
         self.funcTypeInterner = funcTypeInterner
@@ -849,7 +852,8 @@ struct InstructionTranslator: InstructionVisitor {
         self.valueStack = ValueStack(stackLayout: stackLayout)
         self.locals = Locals(types: type.parameters + locals)
         self.functionIndex = functionIndex
-        self.intercepting = intercepting
+        self.isIntercepting = isIntercepting
+        self.isDebugging = isDebugging
         self.constantSlots = ConstSlots(stackLayout: stackLayout)
         self.validator = InstructionValidator(context: module)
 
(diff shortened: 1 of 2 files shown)


JDevlieghere:
FWIW, LLDB and other runtimes (like WARM) expect/use unreachable (0x00, unconditional trap) for breakpoints. What's the motivation for using 0xFF instead?

Utilities/Sources/WasmGen.swift
```diff
@@ -252,7 +258,9 @@ enum WasmGen {
             code += " }\n"
         }
 
-        code += "}"
+        code += """
+        }
+        """
```
MaxDesiatov:
```suggestion
        code += "}"
```

Utilities/Sources/WasmGen.swift
```diff
@@ -527,8 +535,8 @@ enum WasmGen {
         protocol BinaryInstructionDecoder {
             /// Claim the next byte to be decoded
             @inlinable func claimNextByte() throws -> UInt8
-            /// Visit unknown instruction
-            @inlinable func visitUnknown(_ opcode: [UInt8]) throws
+
+            func throwUnknown(_ opcode: [UInt8]) throws -> Never
```
MaxDesiatov, line 538:
```suggestion
            /// Throw an error due to unknown opcode.
            func throwUnknown(_ opcode: [UInt8]) throws -> Never
```

Sources/WasmKit/Translator.swift
```diff
@@ -2257,6 +2257,16 @@ struct InstructionTranslator<Context: TranslatorContext>: InstructionVisitor {
             return .tableSize(Instruction.TableSizeOperand(tableIndex: table, result: LVReg(result)))
         }
     }
+
+    mutating func visitUnknown(_ opcode: [UInt8]) throws -> Bool {
+        guard opcode.count == 1 && opcode[0] == 0xFF else {
```
kateinoigakukun: Could you gate this behind a Translator flag so that it won't accidentally interpret unrecognized opcode when the debugging mode is not enabled for the function?

MaxDesiatov: Added `isDebugging` property and initializer parameter here, which is currently set to `false`.

kateinoigakukun (approved):

PR #206, opened 2025-10-10
Fix errors when building WasmKit for WASI

MaxDesiatov:
Wasm is the only unsupported platform for WasmKit in [the Swift Package Index build matrix](https://swiftpackageindex.com/swiftwasm/WasmKit).

One possible use case is using features like WAT assembler or component model tools in the browser.

Another use case could be polyfilling a future Wasm feature running with older Wasm features, when running WasmKit itself in an older Wasm engine.

For proper testing we also need to fix `swift test` in the Swift toolchain.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -123,7 +123,7 @@ let package = Package(
 
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
-        .package(url: "https://github.com/apple/swift-argument-parser", branch: "release/1.6.2"),
+        .package(url: "https://github.com/apple/swift-argument-parser", from: "1.5.1"),
         .package(url: "https://github.com/apple/swift-system", from: "1.5.0"),
     ]
 } else {


Package.swift
```diff
@@ -121,7 +121,7 @@ let package = Package(
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.2.2"),
-        .package(url: "https://github.com/apple/swift-system", .upToNextMajor(from: "1.3.0")),
+        .package(url: "https://github.com/apple/swift-system", from: "1.6.0"),
```
kateinoigakukun: What's the latest swift-system version in the swiftlang/swift?

MaxDesiatov: That's 1.5.0, so I'll lower it to that version.

Sources/WasmKit/Engine.swift
```diff
@@ -34,10 +34,12 @@ public struct EngineConfiguration {
...
-        /// Direct threaded code
-        /// - Note: This is the default model for platforms that support
-        /// `musttail` calls.
-        case direct
+        #if !os(WASI)
+            /// Direct threaded code
+            /// - Note: This is the default model for platforms that support
+            /// `musttail` calls.
+            case direct
+        #endif
```
kateinoigakukun: We should not conditionize declaration as much as possible because it leads to have huge amount of transitive conditionalization

Package.swift
```diff
@@ -120,8 +123,8 @@ let package = Package(
 
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
-        .package(url: "https://github.com/apple/swift-argument-parser", from: "1.2.2"),
-        .package(url: "https://github.com/apple/swift-system", .upToNextMajor(from: "1.3.0")),
+        .package(url: "https://github.com/apple/swift-argument-parser", branch: "release/1.6.2"),
```
MaxDesiatov: This builds, but we need to wait for a tag created off that branch.

MaxDesiatov: 1.6.2 is tagged now, I lowered the required version to be the same as the Swift toolchain version. WASI build will pick up 1.6.2 automatically as the highest satisfying `from:` constraint.

PR #211, opened 2025-10-21
Add CI job to check auto-generated code is committed

MaxDesiatov:
Added a job to verify that auto-generated code is committed after making changes to a corresponding source of truth.

This should prevent problems, like the recent missing atomic store instructions were committed, but missing in `instructions.json` and vice versa.


.github/workflows/main.yml
```diff
@@ -10,6 +10,23 @@ concurrency:
...
+          fetch-depth: 2
+      - name: Run auto-generated code checks
+        run: |
+          swift run -q --package-path Utilities WasmKitDevUtils wasmgen >> wasmgen.txt
+          swift run -q --package-path Utilities WasmKitDevUtils vmgen >> vmgen.txt
+          cat wasmgen.txt
+          cat vmgen.txt
+          test -s wasmgen.txt && echo '⚠️ Uncommitted changes made by `WasmKitDevUtils wasmgen`' && exit 1
+          test -s vmgen.txt && echo '⚠️ Uncommitted changes made by `WasmKitDevUtils vmgen`' && exit 1
+
```
kateinoigakukun: I'd recommend using the GHA's error reporting syntax like 
https://github.com/swiftwasm/WasmKit/blob/main/.github/workflows/format.yml

PR #212, opened 2025-10-22
GDB RP: support `qWasmCallStack` and debugger breakpoints

MaxDesiatov:
Includes basic tests for breakpoints.

Android CI failure fixed in https://github.com/swiftwasm/WasmKit/pull/216.

Fixing `main` snapshot CI failures depends on an inclusion of https://github.com/swiftlang/swift/commit/b219d4089c922ceb8b700424236ca97f6087a9a1 in a subsequent snapshot.


Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,245 @@
+#if WasmDebuggingSupport
+
+    extension [Int] {
+        /// Uses search to find for an element in an array that's the next closest to a given value.
```
```suggestion
        /// Uses binary search to find an element in `self` that's next closest to a given value.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,245 @@
+#if WasmDebuggingSupport
+
+    extension [Int] {
+        /// Uses binary search to find an element in `self` that's next closest to a given value.
+        /// - Parameter value: the array element to search for or to use as a baseline when searching.
+        /// - Returns: array element `result`, where `result - value` is the smallest possible, and `result > value`.
```
```suggestion
        /// - Returns: array element `result`, where `result - value` is the smallest possible, while
        /// `result > value` also holds.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+                return self[slice.startIndex]
+            }
+        }
+    }
+
+    extension Instance {
+        /// Return an address of WasmKit's iseq bytecode instruction that matches a given Wasm instruction address.
+        /// - Parameter address: the Wasm instruction to find a mapping for.
+        /// - Returns: A tuple with an address of found iseq instruction and the closests matching Wasm instruction
+        /// if no direct match was found.
```
```suggestion
        /// - Returns: A tuple with an address of found iseq instruction and the original Wasm instruction or next 
        /// closest match if no direct match was found.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+                let iseq = handle.wasmToIseqMapping[nextAddress]
+            else {
+                throw Debugger.Error.noInstructionMappingAvailable(address)
+            }
+
+            return (iseq, nextAddress)
+        }
+    }
+
+    /// User-facing debugger state that driven by a debugger host. This implementation has no knowledge of the exact
```
```suggestion
    /// User-facing debugger state driven by a debugger host. This implementation has no knowledge of the exact
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+        private let store: Store
+
+        /// Parsed in-memory representation of a Wasm module instantiated for debugging.
+        private let module: Module
+
+        /// Instance of parsed Wasm ``module``.
+        private let instance: Instance
+
+        /// Reference to the entrypoint function of the currently debugged module, for use in ``stopAtEntrypoint``.
+        private let entrypointFunction: Function
```
MaxDesiatov, line 24:
```suggestion
        /// Reference to the entrypoint function of the currently debugged module, for use in ``stopAtEntrypoint``.
        /// Currently assumed to be the WASI command `_start` entrypoint.
        private let entrypointFunction: Function
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+
+        /// Threading model of the Wasm engine configuration cached for a potentially hot path.
```
```suggestion
        /// Threading model of the Wasm engine configuration, cached for a potentially hot path.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        /// Sets a breakpoint at the first instruction in the entrypoint function of the module instantiated by
+        /// this debugger.
+        package mutating func stopAtEntrypoint() throws {
+            try self.enableBreakpoint(address: self.originalAddress(function: entrypointFunction))
+        }
+
+        /// Finds a Wasm address for the first instruction in a given function.
+        /// - Parameter function: the Wasm function to find a Wasm address for.
```
```suggestion
        /// - Parameter function: the Wasm function to find the first Wasm instruction address for.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+                return wasm.originalAddress
+            case .uncompiled:
+                try function.handle.wasm.ensureCompiled(store: StoreRef(self.store))
+                return try self.originalAddress(function: function)
+            case .compiled:
+                fatalError()
+            }
+        }
+
+        /// Enable a breakpoint at a given Wasm address.
```
```suggestion
        /// Enables a breakpoint at a given Wasm address.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,246 @@
...
+        /// - Parameter address: byte offset of the Wasm instruction that will be replaced with a breakpoint. If no
+        /// direct internal bytecode matching instruction is found, the next closest internal bytecode instruction
+        /// is replaced with a breakpoint. The original instruction to be restored is preserved in debugger state.
```
```suggestion
        /// is replaced with a breakpoint. The original instruction to be restored is preserved in debugger state
        /// represented by `self`.
```

Sources/WasmKit/Execution/Execution.swift
```diff
@@ -315,6 +318,41 @@ func executeWasm(
 }
 
 extension Execution {
+
+    #if WasmDebuggingSupport
+
+        mutating func executeWasm(
```
MaxDesiatov, line 326:
```suggestion
        /// Counterpart to the free `executeWasm` function but implemented as a method of `Execution`,
        /// Useful for representation of debugger state that needs to own `Execution`'s memory.
        mutating func executeWasm(
```

Sources/WasmKit/Execution/Instances.swift
```diff
@@ -83,6 +83,19 @@ struct InstanceEntity /* : ~Copyable */ {
     var functionRefs: Set<InternalFunction>
     var features: WasmFeatureSet
     var dataCount: UInt32?
+    var isDebuggable: Bool
+
+    /// Mapping from iSeq Pc to instruction addresses in the original binary.
```
MaxDesiatov:
```suggestion
    /// Mapping from iseq Pc to instruction addresses in the original binary.
```

Sources/WasmKit/Execution/Instances.swift
```diff
@@ -83,6 +83,19 @@ struct InstanceEntity /* : ~Copyable */ {
...
+    /// Used for handling current call stack requests issued by a ``Debugger`` instance.
+    var iseqToWasmMapping = [Pc: Int]()
+
+    /// Mapping from Wasm instruction addresses in the original binary to iSeq instruction addresses.
```
MaxDesiatov:
```suggestion
    /// Mapping from Wasm instruction addresses in the original binary to iseq instruction addresses.
```

Sources/WasmKit/Module.swift
```diff
@@ -138,9 +138,23 @@ public struct Module {
         Instance(handle: try self.instantiateHandle(store: store, imports: imports), store: store)
     }
 
+    #if WasmDebuggingSupport
+        /// Instantiate this module in the given imports.
```
MaxDesiatov:
```suggestion
        /// Instantiate this module with the given imports.
```

.github/workflows/main.yml
```diff
@@ -233,7 +235,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run Tests on Android emulator
-        uses: skiptools/swift-android-action@v2
+        uses: skiptools/swift-android-action@v2.6.4
```
MaxDesiatov: FYI @marcprux there seems to be a regression with the latest version of the action using swift.org Swift SDKs.

```suggestion
        # For some reason `link-secret-dir-b/secret-c.txt` started throwing `.ELOOP` instead of `.ENOTDIR` in https://github.com/swiftwasm/WasmKit/actions/runs/18847983546/job/53777014067
        # for `pathOpen` test case in `WASITests.swift` when using `v2.7.0`
        uses: skiptools/swift-android-action@v2.6.4
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,248 @@
...
+                    }
+                }
+
+                return self[slice.startIndex]
+            }
+        }
+    }
+
+    extension Instance {
+        /// Return an address of WasmKit's iseq bytecode instruction that matches a given Wasm instruction address.
```
```suggestion
        /// Computes an address of WasmKit's iseq bytecode instruction that matches a given Wasm instruction address.
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -0,0 +1,248 @@
...
+
+            self.breakpoints[wasm] = nil
+            iseq.pointee = oldCodeSlot
+        }
+
+        /// If the module instantiated by the debugger is stopped at a breakpoint, the breakpoint is disabled
+        /// and execution is resumed until the next breakpoint is triggered or all remaining instructions are
+        /// executed. If the module is not stopped at a breakpoint, this function retusn immediately.
+        /// - Returns: `[Value]` result of `entrypointFunction` if current instance ran to completion,
+        /// `nil` if it stopped at a breakpoint.
```
```suggestion
        /// Resumes the module instantiated by the debugger stopped at a breakpoint. The breakpoint is disabled
        /// and execution is resumed until the next breakpoint is triggered or all remaining instructions are
        /// executed. If the module is not stopped at a breakpoint, this function returns immediately.
        /// - Returns: `[Value]` result of `entrypointFunction` if current instance ran to completion,
        /// `nil` if it stopped at a breakpoint.
```

.github/workflows/main.yml
marcprux: Well that's mystifying. I did indeed update the `swift-android-action`, but only to add support for the `nightly-main` option for [swift-version](https://github.com/skiptools/swift-android-action?tab=readme-ov-file#swift-versions) for testing the official Swift SDK for Android builds. The default value for that setting should continue to use the same Swift 6.2 build as before, so there shouldn't be any new breakage.

But it is possible that I somehow introduced some bug or change in the environment. Is it also possible that this is a spurious or flaky error? Did you hit the error multiple times?

BTW, if you are feeling adventurous, I encourage you to add testing against `nightly-main` as well as `6.2` (maybe using a `matrix`) … that'd help us shake out any errors that arise from the new Swift SDK builds.

MaxDesiatov: Yes, I did it hit it multiple times in a row. Haven't tried checking different Swift versions yet.

marcprux: Oh wait, one other change made was that the default Android emulator configuration has [changed](https://github.com/skiptools/swift-android-action/commit/b096effb4d734ee58e77d37e8046e2bce158c8c4#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6L85) from API 30 to API 28 (to match the minimum supported version of the Swift SDK for Android).

I bet _that's_ the issue. You could check this by manually setting `android-api-level: 30` in the action's parameters. 

Of course, it's still an issue if Android is raising a different error for older API levels, but that would be something that would need to be tolerated or worked around.

MaxDesiatov: Thanks! Trying that out in a separate PR https://github.com/swiftwasm/WasmKit/pull/216

Sources/WasmKit/DebuggerInstructionMapping.swift
```diff
@@ -0,0 +1,86 @@
+
+/// Two-way mapping between Wasm and internal iseq bytecode instructions. The implementation of the mapping
+/// is private and is empty when `WasmDebuggingSupport` package trait is disabled.
+struct DebuggerInstructionMapping {
```
MaxDesiatov: Nice side-effect of moving to a separate type is that it's easier to switch to an empty implementation of the mapping based on package traits.

@kateinoigakukun I hope this addresses your review feedback, let me know if anything else is needed to move this forward.

kateinoigakukun (approved):

PR #215, opened 2025-10-27
GDB RP: handle instruction stepping debugger host command

MaxDesiatov:
This change implements [`vCont;s` debugger host command](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#vCont-packet), which allows stepping over Wasm instructions.

This required fixing an off-by-one error in instruction mapping search logic, with new test expectations added in `DebuggerTests/binarySearch` function.

Note that this doesn't handle branching instructions yet, we just increment Pc by one instruction without proper analysis of potential branching (marked as `TODO`).

Corresponding test expectations for the debugger breakpoints and stepping are added in `DebuggerTests/breakpoints` (previously called `DebuggerTests/stopAtEntrypoint`).

### DIFF
--- a/Tests/WasmKitTests/DebuggerTests.swift
+++ b/Tests/WasmKitTests/DebuggerTests.swift
@@ -31,13 +31,15 @@
 
             #expect(try debugger.run() == nil)
 
-            var expectedPc = try #require(debugger.breakpoints.keys.first)
-            #expect(debugger.currentCallStack == [expectedPc])
+            let firstExpectedPc = try #require(debugger.breakpoints.keys.first)
+            #expect(debugger.currentCallStack == [firstExpectedPc])
 
             try debugger.step()
             #expect(try debugger.breakpoints.count == 1)
-            expectedPc = try #require(debugger.breakpoints.keys.first)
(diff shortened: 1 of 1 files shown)


Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -178,6 +179,20 @@
...
+        /// Steps by a single Wasm instruction in the module instantiated by the debugger stopped at a breakpoint.
+        /// The current breakpoint is disabled and new breakpoints are put on the next instruction (or instructions in case
+        /// of multiple possible execution branches). After breakpoints setup, execution is resumed until suspension.
+        /// If the module is not stopped at a breakpoint, this function returns immediately.
+        package mutating func step() throws {
+            guard let currentBreakpoint else {
+                return
+            }
+
+            try self.enableBreakpoint(address: currentBreakpoint.wasmPc + 1)
```
MaxDesiatov, line 192:
```suggestion
            // TODO: analyze actual instruction branching to set the breakpoint correctly.
            try self.enableBreakpoint(address: currentBreakpoint.wasmPc + 1)
```

kateinoigakukun (approved):

PR #222, opened 2025-10-31
Fix multiple bugs in GDB RP functionality

MaxDesiatov:
Multiple issues with GDB RP implementation are addressed here:
1. WasmKit not shutting down after debugger client sends a kill packet. We have to use a wrapping `withThrowingTaskGroup` for existing `withThrowingDiscardingTaskGroup` to propagate debugger handler cancellation error.
2. `isNoAckModeActive` set too early in `GDBHostCommandDecoder`, which led to non-deterministic issues with command decoding.
3. Separate `enum BreakpointState` within `Debugger` to properly express its state machine nature and correctly report exit codes of WASI  modules to the debugger client.
4. Use of `Execution.CurrentMemory.mayUpdateCurrentInstance` without the `from:` parameter, which means we no longer go through an early return path and now correctly set linear memory instance when resuming from a breakpoint.
5. Fixed OOB in binary search implementation in `DebuggerInstructionMapping`, where we had to peek outside of an array slice and use Array's `self` instead of local `slice`.

### DIFF
--- a/Sources/CLI/DebuggerServer.swift
+++ b/Sources/CLI/DebuggerServer.swift
@@ -109,7 +109,7 @@
                         }
                     }
 
-                    // The stream isn't really sending data, just a single type of empty signal, wait for the first one.
+                    // The stream isn't really sending data, just a single empty value, wait for the first one.
                     await shutDownStream.first { _ in true }
                     cancellableGroup.cancelAll()
                 }


Sources/GDBRemoteProtocol/GDBHostCommandDecoder.swift
```diff
@@ -122,6 +118,11 @@ package struct GDBHostCommandDecoder: ByteToMessageDecoder {
             }
 
             startDelimiter = secondStartDelimiter
+
+            if self.isNoAckModeRequested {
+                self.isNoAckModeActive = true
+            }
+
```
MaxDesiatov:
```suggestion
```

Sources/CLI/DebuggerServer.swift
```diff
@@ -71,29 +71,47 @@
...
                                         }
                                     }
-                                } catch {
-                                    logger.error("Error in GDB remote protocol connection channel", metadata: ["error": "\(error)"])
                                 }
                             }
                         }
                     }
+
+                    // The stream isn't really sending data, just a single type of empty signal, wait for the first one.
```
MaxDesiatov:
```suggestion
                    // The stream isn't really sending data, just a single empty value, wait for the first one.
```

PR #225, opened 2025-11-04
GDB RP: breakpoint commands, fix breakpoints across function calls

MaxDesiatov:
This allows setting breakpoints at arbitrary line or function in LLDB CLI interface.

Also fixed multiple bugs:
1. Attempts to set breakpoints for any function other than the entrypoint failed as those were uncompiled functions when lazy compilation is used.
2. WasmKit used wrong order of call stack Pc addresses when sending `qWasmCallStack` response back via GDB RP.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -45,7 +45,7 @@
 
         private var pc = Pc.allocate(capacity: 1)
 
-        /// Addresses of functions in the original Wasm binary, used looking up functions when a breakpoint
+        /// Addresses of functions in the original Wasm binary, used for looking up functions when a breakpoint
         /// is enabled at an arbitrary address if it isn't present in ``InstructionMapping`` yet (i.e. the
         /// was not compiled yet in lazy compilation mode).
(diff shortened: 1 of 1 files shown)


Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -46,6 +45,11 @@
 
         private var pc = Pc.allocate(capacity: 1)
 
+        /// Addresses of functions in the original Wasm binary, used looking up functions when a breakpoint
```
```suggestion
        /// Addresses of functions in the original Wasm binary, used for looking up functions when a breakpoint
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -79,7 +91,7 @@
         /// Finds a Wasm address for the first instruction in a given function.
         /// - Parameter function: the Wasm function to find the first Wasm instruction address for.
         /// - Returns: byte offset of the first Wasm instruction of given function in the module it was parsed from.
-        private func originalAddress(function: Function) throws -> Int {
+        package func originalAddress(function: Function) throws -> Int {
```
```suggestion
        private func originalAddress(function: Function) throws -> Int {
```

PR #226, opened 2025-11-05
GDB RP: add support for reading Wasm locals

MaxDesiatov:
This change fully implements `qWasmLocal` GDB RP host command per the specification updated in https://github.com/llvm/llvm-project/pull/170393.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -87,7 +87,6 @@
             self.store = store
             self.execution = Execution(
                 store: StoreRef(store),
-                stackStart: .init(start: valueStack, count: limit),
                 stackEnd: valueStack.advanced(by: limit)
             )
             self.threadingModel = store.engine.configuration.threadingModel
--- a/Sources/WasmKit/Execution/Execution.swift
+++ b/Sources/WasmKit/Execution/Execution.swift
@@ -8,9 +8,6 @@ struct Execution: ~Copyable {
     /// The reference to the ``Store`` associated with the execution.
     let store: StoreRef
 
-    /// The start of the VM stack space, used for debugging purposes.
-    let stackStart: UnsafeBufferPointer<StackSlot>
-
     /// The end of the VM stack space.
     private let stackEnd: UnsafeMutablePointer<StackSlot>
     /// The error trap thrown during execution.
@@ -19,9 +16,8 @@ struct Execution: ~Copyable {
     private var trap: (error: UnsafeRawPointer, sp: Sp)? = nil
 
     #if WasmDebuggingSupport
-        package init(store: StoreRef, stackStart: UnsafeBufferPointer<StackSlot>, stackEnd: UnsafeMutablePointer<StackSlot>) {
+        package init(store: StoreRef, stackEnd: UnsafeMutablePointer<StackSlot>) {
             self.store = store
-            self.stackStart = stackStart
             self.stackEnd = stackEnd
         }
     #endif
@@ -37,7 +33,7 @@ struct Execution: ~Copyable {
         defer {
             valueStack.deallocate()
         }
-        var context = Execution(store: store, stackStart: .init(start: valueStack, count: limit), stackEnd: valueStack.advanced(by: limit))
+        var context = Execution(store: store, stackEnd: valueStack.advanced(by: limit))
         return try body(&context, valueStack)
     }
 
--- a/Sources/WasmKitGDBHandler/DebuggerMemoryView.swift
+++ b/Sources/WasmKitGDBHandler/DebuggerMemoryView.swift
@@ -15,7 +15,7 @@
     import NIOCore
     import WasmKit
 
-    package struct DebuggerMemoryCache: ~Copyable {
+    package struct DebuggerMemoryView: ~Copyable {
         package static let executableCodeOffset = UInt64(0x4000_0000_0000_0000)
 
         private let allocator: ByteBufferAllocator
--- a/Sources/WasmKitGDBHandler/WasmKitGDBHandler.swift
+++ b/Sources/WasmKitGDBHandler/WasmKitGDBHandler.swift
@@ -56,7 +56,7 @@
         private let allocator: ByteBufferAllocator
         private var debugger: Debugger
 
-        private var memoryCache: DebuggerMemoryCache
+        private var memoryView: DebuggerMemoryView
 
         package init(
             moduleFilePath: FilePath,
@@ -85,7 +85,7 @@
                 throw Error.stoppingAtEntrypointFailed
             }
 
-            self.memoryCache = DebuggerMemoryCache(allocator: allocator, wasmBinary: wasmBinary)
+            self.memoryView = DebuggerMemoryView(allocator: allocator, wasmBinary: wasmBinary)
         }
 
         private func hexDump<I: FixedWidthInteger>(_ value: I, endianness: Endianness) -> String {
@@ -117,7 +117,7 @@
                 switch self.debugger.state {
                 case .stoppedAtBreakpoint(let breakpoint):
                     let pc = breakpoint.wasmPc
-                    let pcInHostAddressSpace = UInt64(pc) + DebuggerMemoryCache.executableCodeOffset
+                    let pcInHostAddressSpace = UInt64(pc) + DebuggerMemoryView.executableCodeOffset
                     result.append(("thread-pcs", self.hexDump(pcInHostAddressSpace, endianness: .big)))
                     result.append(("00", self.hexDump(pcInHostAddressSpace, endianness: .little)))
                     result.append(("reason", "trace"))
@@ -221,7 +221,7 @@
                         """
                         l<library-list>\
                         <library name="\(self.moduleFilePath.string)">\
-                        <section address="0x\(String(DebuggerMemoryCache.executableCodeOffset, radix: 16))"/>\
+                        <section address="0x\(String(DebuggerMemoryView.executableCodeOffset, radix: 16))"/>\
                         </library>\
                         </library-list>
                         """)
@@ -238,7 +238,7 @@
                 else { throw Error.unknownReadMemoryArguments }
 
                 responseKind = .hexEncodedBinary(
-                    try self.memoryCache.readMemory(
+                    try self.memoryView.readMemory(
                         debugger: self.debugger,
                         addressInProtocolSpace: addressInProtocolSpace,
                         length: length
@@ -249,7 +249,7 @@
                 let callStack = self.debugger.currentCallStack
                 var buffer = self.allocator.buffer(capacity: callStack.count * 8)
                 for pc in callStack {
-                    buffer.writeInteger(UInt64(pc) + DebuggerMemoryCache.executableCodeOffset, endianness: .little)
+                    buffer.writeInteger(UInt64(pc) + DebuggerMemoryView.executableCodeOffset, endianness: .little)
                 }
                 responseKind = .hexEncodedBinary(buffer.readableBytesView)
 
@@ -288,7 +288,7 @@
                             argumentsString: command.arguments,
                             separator: ",",
                             endianness: .big
-                        ) - DebuggerMemoryCache.executableCodeOffset)
+                        ) - DebuggerMemoryView.executableCodeOffset)
                 )
                 responseKind = .ok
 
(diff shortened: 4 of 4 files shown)


Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -43,13 +46,16 @@
 
         package private(set) var state: State
 
-        private var pc = Pc.allocate(capacity: 1)
+        /// Pc ofthe final instruction that a successful program will execute, initialized with `Instruction.endofExecution`
+        private let endOfExecution = Pc.allocate(capacity: 1)
 
-        /// Addresses of functions in the original Wasm binary, used for looking up functions when a breakpoint
+        /// Addresses of functions in the original Wasm binary, used looking up functions when a breakpoint
```
```suggestion
        /// Addresses of functions in the original Wasm binary, used for looking up functions when a breakpoint
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -43,13 +46,16 @@
...
         /// is enabled at an arbitrary address if it isn't present in ``InstructionMapping`` yet (i.e. the
         /// was not compiled yet in lazy compilation mode).
         private let functionAddresses: [(address: Int, instanceFunctionIndex: Int)]
 
+        private var stackFrame = DebuggerStackFrame()
```
```suggestion
        /// Cache for stack frames that use debugger-friendly memory layout.
        private var stackFrame = DebuggerStackFrame()
```

.github/workflows/main.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             wasi-swift-sdk-download: "https://download.swift.org/development/wasm-sdk/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a_wasm.artifactbundle.tar.gz"
             wasi-swift-sdk-id: swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a_wasm
             wasi-swift-sdk-checksum: "b64dfad9e1c9ccdf06f35cf9b1a00317e000df0c0de0b3eb9f49d6db0fcba4d9"
-            test-args: "--sanitize address --traits WasmDebuggingSupport"
```
kateinoigakukun, line 48: Do we no longer test debugger on CI?

Sources/WasmKit/Execution/DebuggerStackFrame.swift
```diff
@@ -0,0 +1,98 @@
+#if WasmDebuggingSupport
+
+    package struct DebuggerStackFrame: ~Copyable {
```
MaxDesiatov:
```suggestion
    /// Encapsulates a caching buffer that stores a stack frame in debugger-friendly
    /// memory layout that's distinct from WasmKit's iseq stack frame memory layout.
    package struct DebuggerStackFrame: ~Copyable {
```

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -43,13 +46,17 @@
 
         package private(set) var state: State
 
-        private var pc = Pc.allocate(capacity: 1)
+        /// Pc ofthe final instruction that a successful program will execute, initialized with `Instruction.endofExecution`
+        private let endOfExecution = Pc.allocate(capacity: 1)
```
kateinoigakukun: We don't need to allocate it on heap, we can just stack alloc it?

```swift
var endOfExecution: Instruction = Instruction.endOfExecution.headSlot(threadingModel: threadingModel)
let result = try self.execution.executeWasm(
    threadingModel: self.threadingModel,
    function: self.entrypointFunction.handle,
    type: self.entrypointFunction.type,
    arguments: [],
    sp: self.valueStack,
    pc: &endOfExecution
)
```

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
```diff
@@ -0,0 +1,108 @@
...
+                    throw Debugger.Error.stackLocalIndexOOB(localIndex)
+                }
+
+                frameBase = base
+                frameLayout = layout
+            } else {
+                let (base, layout) = try debugger.packedStackFrame(frameIndex: frameIndex) { span, layout in
+                    let baseAddress = self.stackFrames.writerIndex
+                    self.stackFrames.writeBytes(span)
+                    print("written to stackFrames: \(self.stackFrames.hexDump(format: .plain))")
```
kateinoigakukun: Let's remove debug prints

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
```diff
@@ -0,0 +1,108 @@
...
+
+            if addressInProtocolSpace >= self.stackOffsetInProtocolSpace {
+                let stackAddress = Int(addressInProtocolSpace - self.stackOffsetInProtocolSpace)
+                if stackAddress + length > self.stackFrames.readableBytes {
+                    length = self.stackFrames.readableBytes - stackAddress
+                }
+
+                return self.stackFrames.readableBytesView[stackAddress..<(stackAddress + length)]
+            } else if addressInProtocolSpace >= Self.executableCodeOffset {
+                print("wasmBinary")
```
kateinoigakukun: ditto

Sources/WasmKit/Execution/DebuggerStackFrame.swift
```diff
@@ -0,0 +1,100 @@
...
+        private var buffer = UnsafeMutableRawBufferPointer.allocate(byteCount: 1, alignment: 8)
+
+        package struct Layout {
+            /// Array of byte offsets from frame base address for each local at a given index.
+            package fileprivate(set) var localOffsets = [Int]()
+        }
+
+        init() {
+            buffer.initializeMemory(as: Int.self, repeating: 0)
```
kateinoigakukun:
```swift
var buffer = UnsafeMutableRawBufferPointer.allocate(byteCount: 1, alignment: 8)
buffer.initializeMemory(as: Int.self, repeating: 0)
```

Doesn't this cause buffer over write?

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
```diff
@@ -0,0 +1,108 @@
...
+                guard layout.localOffsets.count > localIndex else {
+                    throw Debugger.Error.stackLocalIndexOOB(localIndex)
+                }
+
+                frameBase = base
+                frameLayout = layout
+            } else {
+                let (base, layout) = try debugger.packedStackFrame(frameIndex: frameIndex) { span, layout in
+                    let baseAddress = self.stackFrames.writerIndex
+                    self.stackFrames.writeBytes(span)
```
kateinoigakukun: ~~What happens if we receive commands in the following order?~~
```
wasmLocal(frame: 0, local: 0) ==> 0x4000000000001000
wasmLocal(frame: 1, local: 0) ==> 0x4000000000002000
readMemory(0x4000000000001000)
```

~~IMHO it's better to start from a stateless approach for simplicity, given that the number of stack frames is very small.~~

Update: Never mind, the cached is distinguished by frame index.

.github/workflows/main.yml
MaxDesiatov: This is Swift 6.1 job, which can no longer support the debugger functionality due to use of `Span` introduced in this PR.

Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -43,13 +46,17 @@
 
         package private(set) var state: State
 
-        private var pc = Pc.allocate(capacity: 1)
+        /// Pc ofthe final instruction that a successful program will execute, initialized with `Instruction.endofExecution`
```
```suggestion
        /// Pc of the final instruction that a successful program will execute, initialized with `Instruction.endofExecution`
```

.github/workflows/main.yml
kateinoigakukun: Ah, ok. We still test 6.2 with debugger support

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
```diff
@@ -0,0 +1,108 @@
...
+        package init(allocator: ByteBufferAllocator, wasmBinary: ByteBuffer) {
+            self.allocator = allocator
+
+            var stackOffset = Int(Self.executableCodeOffset) + wasmBinary.readableBytes
+            // Untyped raw Wasm values in VM's stack are stored as `UInt64`.
+            stackOffset.roundUpToAlignment(for: UInt64.self)
+
+            self.stackOffsetInProtocolSpace = UInt64(stackOffset)
+            self.wasmBinary = wasmBinary
+            self.stackFrames = allocator.buffer(capacity: 0)
```
kateinoigakukun: I think it's better to simply use Optional here instead of allocating 0 so that we don't need to remember about `malloc(0)` semantics 😅

MaxDesiatov: Interesting, what's that `malloc(0)` semantics?

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
kateinoigakukun: On second thought, I still feel it's better to start with a stateless approach because manually invalidating the cache is pretty error-prone. I think we can start from a simpler "view" without any state and revisit the caching option when we really have performance issues.

MaxDesiatov: Yes, I don't like this approach, but after some investigation, I have a suspicion that `qWasmLocal` GDB RP command should return the actual local instead of some fake memory address to a buffer where locals are stored, based on how other Wasm runtimes implemented it. In that case, current documentation that says it should return an address is incorrect, but then we probably can remove `DebuggerMemoryCache` altogether. https://lldb.llvm.org/resources/lldbgdbremote.html#qwasmlocal

I need a bit more time to confirm this though. Will mark this PR as draft until then.

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
kateinoigakukun: Section 7.20.3 Memory management functions in C99 spec says:

> If the size of the space requested is zero, the behavior is implementation defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf

So `free(malloc(0))` is well-defined without any leak but `malloc(0)` alone may leak something.

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
kateinoigakukun: If we just need to return the value directly, it simplifies a lot! Let's check LLDB side and other implementations

MaxDesiatov: Indeed, stack values are returned directly, which simplifies this a lot and I hope addresses most of your feedback. I've updated extension docs here https://github.com/llvm/llvm-project/pull/170393

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
```diff
@@ -0,0 +1,57 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if WasmDebuggingSupport
+
+    import NIOCore
+    import WasmKit
+
+    package struct DebuggerMemoryCache: ~Copyable {
```
kateinoigakukun: This type no longer has any state or cache, so let's call it `DebuggerMemoryView`.

Sources/WasmKit/Execution/Execution.swift
```diff
@@ -7,16 +7,21 @@ import _CWasmKit
 struct Execution: ~Copyable {
     /// The reference to the ``Store`` associated with the execution.
     let store: StoreRef
+
+    /// The start of the VM stack space, used for debugging purposes.
+    let stackStart: UnsafeBufferPointer<StackSlot>
+
```
kateinoigakukun: Can you double check if we really need to add this field here? It's important to keep the `Execution` type to be <= 4 words to pass/return it directly on function calls. If we exceed that threshold, `LoadableByAddress` will demote it to pass/return indirectly via stack-allocs.

MaxDesiatov: Thanks, that's cleaned up.

Sources/WasmKitGDBHandler/DebuggerMemoryCache.swift
MaxDesiatov: Good point, that's addressed now.

PR #230, opened 2025-11-26
WASI MemoryFileSystem

andrewmd5:
Adds an in-memory file system to the WASI bridge (with the ability to mount actual file descriptors.)

### DIFF
--- a/Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
+++ b/Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
@@ -1,5 +1,30 @@
 import SystemPackage
 
+#if canImport(Darwin)
+    import Darwin
+#elseif canImport(Glibc)
+    import Glibc
+#elseif canImport(Musl)
+    import Musl
+#elseif os(Windows)
+    import ucrt
+    import WinSDK
+#endif
+
+func currentTimestamp() -> WASIAbi.Timestamp {
+    #if os(Windows)
+        var ft = FILETIME()
+        GetSystemTimeAsFileTime(&ft)
+        let intervals = (Int64(ft.dwHighDateTime) << 32) | Int64(ft.dwLowDateTime)
+        let unixIntervals = intervals - 116_444_736_000_000_000
+        return WASIAbi.Timestamp(unixIntervals * 100)
+    #else
+        var ts = timespec()
+        clock_gettime(CLOCK_REALTIME, &ts)
+        return WASIAbi.Timestamp(ts.tv_sec) * 1_000_000_000 + WASIAbi.Timestamp(ts.tv_nsec)
+    #endif
+}
+
 /// Base protocol for all file system nodes in memory.
 protocol MemFSNode: AnyObject {
     var type: MemFSNodeType { get }
@@ -17,7 +42,41 @@ final class MemoryDirectoryNode: MemFSNode {
     let type: MemFSNodeType = .directory
     private var children: [String: MemFSNode] = [:]
 
-    init() {}
+    private var _atim: WASIAbi.Timestamp
+    private var _mtim: WASIAbi.Timestamp
+    private var _ctim: WASIAbi.Timestamp
+
+    init() {
+        let now = currentTimestamp()
+        self._atim = now
+        self._mtim = now
+        self._ctim = now
+    }
+
+    var timestamps: (atim: WASIAbi.Timestamp, mtim: WASIAbi.Timestamp, ctim: WASIAbi.Timestamp) {
+        return (_atim, _mtim, _ctim)
+    }
+
+    func touchAccessTime() {
+        _atim = currentTimestamp()
+    }
+
+    func touchModificationTime() {
+        let now = currentTimestamp()
+        _mtim = now
+        _ctim = now
+    }
+
+    func setTimes(atim: WASIAbi.Timestamp?, mtim: WASIAbi.Timestamp?) {
+        let now = currentTimestamp()
+        if let atim = atim {
+            _atim = atim
+        }
+        if let mtim = mtim {
+            _mtim = mtim
+        }
+        _ctim = now
+    }
 
     func getChild(name: String) -> MemFSNode? {
         return children[name]
@@ -25,14 +84,20 @@ final class MemoryDirectoryNode: MemFSNode {
 
     func setChild(name: String, node: MemFSNode) {
         children[name] = node
+        touchModificationTime()
     }
 
     @discardableResult
     func removeChild(name: String) -> Bool {
-        return children.removeValue(forKey: name) != nil
+        let removed = children.removeValue(forKey: name) != nil
+        if removed {
+            touchModificationTime()
+        }
+        return removed
     }
 
     func listChildren() -> [String] {
+        touchAccessTime()
         return Array(children.keys).sorted()
     }
 
@@ -46,8 +111,16 @@ final class MemoryFileNode: MemFSNode {
     let type: MemFSNodeType = .file
     var content: FileContent
 
+    private var _atim: WASIAbi.Timestamp
+    private var _mtim: WASIAbi.Timestamp
+    private var _ctim: WASIAbi.Timestamp
+
     init(content: FileContent) {
         self.content = content
+        let now = currentTimestamp()
+        self._atim = now
+        self._mtim = now
+        self._ctim = now
     }
 
     convenience init(bytes: some Sequence<UInt8>) {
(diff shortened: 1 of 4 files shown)


Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+    case bytes([UInt8])
+    case handle(FileDescriptor)
+}
+
+/// Public protocol for file system providers that users interact with.
+///
+/// This protocol exposes only user-facing methods for managing files and directories.
+public protocol FileSystemProvider {
+    /// Adds a file to the file system with the given byte content.
+    func addFile(at path: String, content: [UInt8]) throws
```
MaxDesiatov: Let's not hardcode byte arrays here, but allow more types, like `ArraySlice<UInt8>`, `ByteBufferView`, `UnsafeBufferPointer<UInt8>` etc.

```suggestion
    func addFile(at path: String, content: some Sequence<UInt8>) throws
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+
+    /// Removes a file from the file system.
+    func removeFile(at path: String) throws
+}
+
+/// Internal protocol for file system implementations used by WASI.
+///
+/// This protocol contains WASI-specific implementation details that should not
+/// be exposed to library users.
+internal protocol FileSystem {
```
MaxDesiatov: `internal` is the default, so an explicit keyword is redundant here. Also, naming currently doesn't make the difference from `FileSystemProvider` obvious at all. Let's make the name at least a bit more distinct, WDYT of `FileSystemImplementation`?

Also, let's allow non-copyable implementations too.

```suggestion
protocol FileSystemImplementation: ~Copyable {
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+    /// Returns the list of pre-opened directory paths.
+    func getPreopenPaths() -> [String]
+
+    /// Opens a directory and returns a WASIDir implementation.
+    func openDirectory(at path: String) throws -> any WASIDir
```
MaxDesiatov, line 166: I wonder if the use of existentials is justified here? Could we use opaque types instead?

```suggestion
    func openDirectory(at path: String) throws -> some WASIDir
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+    func removeFile(at path: String) throws
+}
+
+/// Internal protocol for file system implementations used by WASI.
+///
+/// This protocol contains WASI-specific implementation details that should not
+/// be exposed to library users.
+internal protocol FileSystem {
+    /// Returns the list of pre-opened directory paths.
+    func getPreopenPaths() -> [String]
```
MaxDesiatov: `get`-prefixed functions are not necessarily idiomatic in Swift, I'd suggest converting this to a ready-only property. 

```suggestion
    var preopenPaths: [String] { get }
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+
+    /// Opens a directory and returns a WASIDir implementation.
+    func openDirectory(at path: String) throws -> any WASIDir
+
+    /// Opens a file or directory from a directory file descriptor.
+    func openAt(
+        dirFd: any WASIDir,
```
MaxDesiatov, line 170: Ditto, is an existential required here?

```suggestion
        dirFd: some WASIDir,
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+        path: String,
+        oflags: WASIAbi.Oflags,
+        fsRightsBase: WASIAbi.Rights,
+        fsRightsInheriting: WASIAbi.Rights,
+        fdflags: WASIAbi.Fdflags,
+        symlinkFollow: Bool
+    ) throws -> FdEntry
+
+    /// Creates a standard I/O file entry for stdin/stdout/stderr.
+    func createStdioFile(fd: FileDescriptor, accessMode: FileAccessMode) -> any WASIFile
```
MaxDesiatov, line 180:
```suggestion
    func createStdioFile(fd: FileDescriptor, accessMode: FileAccessMode) -> some WASIFile
```

Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
```diff
@@ -0,0 +1,216 @@
+import SystemPackage
+
+/// Base protocol for all file system nodes in memory.
+internal protocol MemFSNode: AnyObject {
```
MaxDesiatov: Is there a strong reason to require `AnyObject` here? What would exactly classes (as opposed to non-copyable types) bring to the table other than reference counting performance hit?
```suggestion
protocol MemFSNode {
```

Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
```diff
@@ -0,0 +1,216 @@
...
+    var type: MemFSNodeType { get }
+}
+
+/// Types of file system nodes.
+internal enum MemFSNodeType {
```
MaxDesiatov:
```suggestion
enum MemFSNodeType {
```

Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
```diff
@@ -0,0 +1,216 @@
...
+    case directory
+    case file
+    case characterDevice
+}
+
+/// A directory node in the memory file system.
+internal final class MemoryDirectoryNode: MemFSNode {
```
MaxDesiatov:
```suggestion
final class MemoryDirectoryNode: MemFSNode {
```

Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
```diff
@@ -0,0 +1,216 @@
...
+        return Array(children.keys).sorted()
+    }
+
+    func childCount() -> Int {
+        return children.count
+    }
+}
+
+/// A regular file node in the memory file system.
+internal final class MemoryFileNode: MemFSNode {
```
MaxDesiatov:
```suggestion
final class MemoryFileNode: MemFSNode {
```

Sources/WASI/MemoryFileSystem/MemoryFileEntry.swift
```diff
@@ -0,0 +1,309 @@
+import SystemPackage
+
+/// A WASIFile implementation for regular files in the memory file system.
+internal final class MemoryFileEntry: WASIFile {
```
MaxDesiatov:
```suggestion
final class MemoryFileEntry: WASIFile {
```

Sources/WASI/FileSystem.swift
```diff
@@ -120,3 +127,55 @@ struct FdTable {
...
+/// Content of a file that can be retrieved from the file system.
+public enum FileContent {
+    case bytes([UInt8])
+    case handle(FileDescriptor)
+}
+
+/// Public protocol for file system providers that users interact with.
+///
+/// This protocol exposes only user-facing methods for managing files and directories.
+public protocol FileSystemProvider {
```
MaxDesiatov: Let's allow non-copyable implementations here.

```suggestion
public protocol FileSystemProvider: ~Copyable {
```

Sources/WASI/Platform/HostFileSystem.swift
```diff
@@ -0,0 +1,130 @@
...
+    import WASILibc
+#else
+    #error("Unsupported Platform")
+#endif
+
+/// A file system implementation that directly accesses the host operating system's file system.
+///
+/// This implementation provides access to actual files and directories on the host system,
+/// with appropriate sandboxing through pre-opened directories.
+public final class HostFileSystem: FileSystemProvider, FileSystem {
```
MaxDesiatov: Usual concern WRT classes overhead here. IIUC we're not relying on class hierarchies, so why not make this a non-copyable struct to avoid the ARC overhead?

```suggestion
public struct HostFileSystem: FileSystemProvider, FileSystemImplementation, ~Copyable {
```

Sources/WASI/FileSystem.swift
andrewmd5: `WASIDir` seems to be defined as a protocol so I believe an existential is how its used elsewhere (i.e in Directory.swift)

Sources/WASI/MemoryFileSystem/MemoryFSNodes.swift
andrewmd5: Multiple file entries need to share mutable access to the same node (e.g., a file opened twice should see the same content). With classes, this is natural, both entries hold references to the same object. With non-copyable structs, we'd need an indirection layer (handles/indices into a central store), which adds complexity. Open to that refactor if the ref-counting overhead is measurable or if I am mistaken (Swift is arguably very new to me.)

Sources/WASI/Platform/HostFileSystem.swift
andrewmd5: The reason I went with class here is that `FileSystemImplementation` needs to be stored as a property in `WASIBridgeToHost`, and Swift's existential types `(any Protocol)` don't support `~Copyable` types. If we make this a non-copyable struct, we'd need to make `WASIBridgeToHost` generic.

MaxDesiatov: @kateinoigakukun do you have strong opinions about this? If these were non-public APIs I'd be ok in principle, but wouldn't be great to commit to classes and existentials here without providing high-performance alternatives for those who need it.

kateinoigakukun:
I agree with @andrewmd5's point regarding the reality of poor implementation status for Component Model in the ecosystem. So overall, it makes sense to me to add an in-memory FS implementation alongside the native implementation.

On the other hand, I think we need to refine the added APIs in this pull to be more Embedded Swift/performance-wise friendly before landing. At least we can avoid most of the existential container usage. If you don't mind, I can push my proposal to this branch. wdyt @andrewmd5 ?

andrewmd5:
Sounds good to me.

Sent via Superhuman ( ***@***.*** )

kateinoigakukun:
@andrewmd5 It seems I don't have access to push your branch, so would you mind cherry-picking my change to your branch if you are happy with it? https://github.com/swiftwasm/WasmKit/commit/23dc9f7b7712d594a2dfa76347f248a15754f378

kateinoigakukun:
@andrewmd5 Just a gentle ping :) If you are busy, I can re-open a PR with my change on top of your branch and merge it to main.

PR #237, opened 2025-12-08
Fix breakpoints in loops & recursive functions

MaxDesiatov:
Currently, when continuing from a breakpoint, the debugger needs to remove it and restore the original instruction before resuming. The conventionally expected behaviour is for the original breakpoint to be preserved. For that, we don't fully resume, but only step to the next instruction, restore the original breakpoint, and then finally resume again.

This still doesn't account for branching instructions, in the same way that the current instruction stepping does not. To be resolved in a future PR, as that requires precise instruction control flow analysis.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -209,12 +209,12 @@
                         case .token:
                             try self.execution.runTokenThreaded(sp: &sp, pc: &pc, md: &md, ms: &ms)
                         }
-                    } catch is Execution.EndOfExecution {
+                    } catch let end as Execution.EndOfExecution {
                         // The module successfully executed till the "end of execution" instruction.
-                        let type = self.store.engine.funcTypeInterner.resolve(currentFunction.type)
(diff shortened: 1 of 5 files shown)


Sources/WasmKit/Execution/Debugger.swift
```diff
@@ -209,12 +209,12 @@
...
                         }
-                    } catch is Execution.EndOfExecution {
+                    } catch let end as Execution.EndOfExecution {
                         // The module successfully executed till the "end of execution" instruction.
-                        let type = self.store.engine.funcTypeInterner.resolve(currentFunction.type)
+                        let type = self.entrypointFunction.type
                         self.state = .entrypointReturned(
                             type.results.enumerated().map { (i, type) in
-                                sp[VReg(i)].cast(to: type)
+                                end.sp[VReg(i)].cast(to: type)
```
MaxDesiatov, line 217: `.entrypointReturned` state was previously set incorrectly, as we were capturing outdated `sp` instead of getting the actual `sp` from the final "end of execution" instruction.

PR #246, opened 2026-01-02
Add `wat2wasm` CLI command as `func wat2wasm` wrapper

MaxDesiatov:
This is a CLI wrapper around existing `wat2wasm` free function in the public API. In the future we could provide a `wasm2wat` command to mirror this.

### DIFF
--- a/Sources/CLI/CMakeLists.txt
+++ b/Sources/CLI/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_executable(wasmkit-cli
-    Commands/Run.swift
     Commands/Explore.swift
+    Commands/Run.swift
+    Commands/Wat2wasm.swift
     CLI.swift
 )
 


Sources/CLI/Commands/Assemble.swift
```diff
@@ -0,0 +1,99 @@
...
+import SystemPackage
+import WAT
+import WasmKit
+
+struct Assemble: ParsableCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Assemble WebAssembly text into a WebAssembly binary",
+        discussion: """
+            This command parses a file in WebAssembly Text Format (`.wat`),
+            assembles it into a binary and writes the result to a given
```
MaxDesiatov:
```suggestion
            This command parses a file in WebAssembly Text Format (`.wat`), \
            assembles it into a binary and writes the result to a given \
```

Sources/CLI/Commands/Assemble.swift
```diff
@@ -0,0 +1,99 @@
...
+import SystemPackage
+import WAT
+import WasmKit
+
+struct Assemble: ParsableCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Assemble WebAssembly text into a WebAssembly binary",
+        discussion: """
+            This command parses a file in WebAssembly Text Format (`.wat`), \
+            assembles it into a binary and writes the result to a given \
```
MaxDesiatov:
```suggestion
            assembles it into a binary, and writes the result to a given \
```

Sources/CLI/Commands/Assemble.swift
```diff
@@ -0,0 +1,99 @@
...
+import SystemPackage
+import WAT
+import WasmKit
+
+struct Assemble: ParsableCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Assemble WebAssembly text into a WebAssembly binary",
+        discussion: """
+            This command parses a file in WebAssembly Text Format (`.wat`), \
+            assembles it into a binary, and writes the result to a given \
```
MaxDesiatov:
```suggestion
            Parse a file in WebAssembly Text Format (`.wat`), \
            assemble it into a binary, and write the result to a given \
```

Sources/CLI/Commands/Assemble.swift
```diff
@@ -0,0 +1,99 @@
+import ArgumentParser
+import SystemPackage
+import WAT
+import WasmKit
+
+struct Assemble: ParsableCommand {
```
kateinoigakukun: How about calling it `wasmkit wat2wasm foo.wat`? I personally prefer aligning the subcommand name with the widely known `wat2wasm` convention, which helps reduce cognitive overhead for users familiar with existing WebAssembly tooling.

Sources/CLI/Commands/Assemble.swift
```diff
@@ -0,0 +1,99 @@
...
+            outputPath,
+            .writeOnly,
+            options: [.create],
+            permissions: [.ownerReadWrite, .groupRead, .otherRead]
+        )
+        defer { try? outputHandle.close() }
+
+        try outputHandle.writeAll(wasm)
+
+        print("✅ Wasm binary successfully assembled and written to `\(outputPath.string)`.")
```
kateinoigakukun: Also, printing to stdout on success feels a bit undesirable. 
If we want to support piping or `-o -` in the future, stdout should be treated as a data channel only. 
Status or informational messages should either go to stderr or be omitted entirely.

kateinoigakukun (approved):

PR #248, opened 2026-01-05
Add experimental LLVM backend scaffolding

MaxDesiatov:
Depends on https://github.com/swiftwasm/WasmKit/pull/249.

LLVM native codegen backend in the future could provide Wasm execution performance that's closer to native code, and then become a viable alternative or a complementary component to the existing interpreter.

In the meantime, we're adding basic scaffolding here. It can translate a few of the core Wasm spec instructions into LLVM IR, then into native machine code in an object file, finally linked as a dynamic library that WasmKit can `dlopen`/`dlsym` and call into.

As SwiftPM is unable to build LLVM and can't find LLVM headers without hardcoding absolute paths (or without adding side effects and I/O) in package manifests, LLVM backend is only available when building WasmKit with CMake and `-DWASMKIT_BUILD_LLVM_BACKEND=ON` option passed.


kateinoigakukun (commented):
Thanks for working on this, Max. This is exciting!

However, I'd like to raise some concerns about merging this at the current stage. While I understand this is marked as "experimental scaffolding," merging it into main would establish architectural patterns that may be difficult to change later and add non-trivial maintenance costs. Several fundamental design decisions need to be addressed before we commit to this approach:

### 1. Runtime/JIT boundary architecture and calling conventions

The current implementation uses `dlopen`/`dlsym` with C calling convention and manual per-signature wrappers (`U64Args1Result1`, `U64Args2Result1`, etc.). This approach doesn't scale and will need to be replaced.

Eventually, we'll need to:
- JIT-compile thunks for each function signature to handle JIT <-> Runtime transitions
- Define a unified calling convention for these boundaries

The calling convention design is critical because:
- It affects the entire JIT code generation strategy
- A poorly chosen convention will cause performance issues (extra spills, inefficient argument passing)
- Changing it later requires significant refactoring of the code generator

We should settle on this design before committing the basic structure.

### 2. Selective compilation strategy

Not all instructions should be JIT-compiled. Instructions like `memory.*` and `table.*` are complex and maintenance-heavy to implement in LLVM IR. We need a clear strategy for which instructions get compiled vs. fall back to runtime and how we can reuse the existing implementations.

### 3. Public API surface

- Will this be exposed as a Swift API, CLI option, or both?
- What will the user-facing configuration look like?
- How do users opt-in/out of JIT compilation?

Understanding the API surface helps guide internal architecture decisions.

### 4. Security model

JIT compilation introduces new security considerations:
- W^X policy compliance
- Linear memory protection

These should be addressed in the design phase, as it's not easy to retrofit later.

### 5. Maintenance and testing strategy

With two execution paths (interpreter + LLVM backend):
- How do we ensure both implementations stay in sync with the Wasm spec? I think we don't have enough capacity to maintain both engines up to date with the latest spec. We might need a fallback mechanism when a given program uses instructions not supported by the LLVM backend.
- What's the testing strategy for verifying semantic equivalence?

### Suggestion

I propose we continue developing this work in this feature branch until we have consensus on these architectural questions. This allows:
- Rapid iteration without affecting main
- Time to create a design document or RFC addressing these concerns
- Opportunity to prototype different approaches to the runtime boundary

Once the core design is settled, we can merge with confidence that we won't need major refactoring.

Would it make sense to start with a design document that addresses these points? I'm happy to collaborate on that.

PR #249, opened 2026-01-07
NFC: make CLI commands reusable with new `CLICommands` module

MaxDesiatov:
Prerequisite for https://github.com/swiftwasm/WasmKit/pull/248 to make existing CLI commands reused in both `wasmkit` and `wasmkit-llvm` executables.

### DIFF
--- a/Sources/CLICommands/CMakeLists.txt
+++ b/Sources/CLICommands/CMakeLists.txt
@@ -5,4 +5,16 @@ add_wasmkit_library(CLICommands
 )
 
 target_link_wasmkit_libraries(CLICommands PUBLIC
-  ArgumentParser WAT WasmKitWASI)
+  WAT WasmKitWASI)
+
+add_dependencies(
+  CLICommands
+
+  ArgumentParser
+)
+
+target_link_libraries(CLICommands
+
+  PUBLIC
+  ArgumentParser
+)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,8 @@ add_compile_definitions(
 
 include(FetchContent)
 
+option(WASMKIT_BUILD_CLI "Build wasmkit-cli" ON)
+
 if(WASMKIT_BUILD_CLI)
   set(BUILD_TESTING OFF) # disable ArgumentParser tests
   find_package(ArgumentParser CONFIG)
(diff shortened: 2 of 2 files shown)


Sources/CLICommands/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  find_package(ArgumentParser CONFIG)
+  if(NOT ArgumentParser_FOUND)
+    message("-- Vending ArgumentParser")
+    FetchContent_Declare(ArgumentParser
+            GIT_REPOSITORY https://github.com/apple/swift-argument-parser
+            GIT_TAG 1.6.1
+    )
+    FetchContent_MakeAvailable(ArgumentParser)
+  endif()
+endif()
```
kateinoigakukun: Can you keep this FetchContent stuff in the root `CMakeLists.txt`? This is not a module-specific configuration but a project-wide dependency, so I prefer leaving it in the root so that we can find project dependencies easily

Sources/CLICommands/Run.swift
```diff
@@ -311,3 +314,44 @@ struct Run: AsyncParsableCommand {
...
+            try fileHandle.read(fromAbsoluteOffset: 0, into: .init($0))
+        }
+        return try WasmKit.parseWasm(bytes: wat2wasm(wat))
+    } else {
+        return try WasmKit.parseWasm(filePath: filePath)
+    }
+}
+
+extension Run {
+    package static func parseInvocation(arguments: [String]) -> (functionName: String?, parameters: [Value]) {
```
kateinoigakukun, line 320: Did you mean using this method in `instantiateNonWASI`?

MaxDesiatov: Good catch, thanks! That's addressed now

Sources/CLICommands/CMakeLists.txt
MaxDesiatov: The problem then is that I can't specify `ArgumentParser` as a dependency of `CLICommands` if `FetchContent_MakeAvailable(ArgumentParser)` is not in the same `CMakeLists.txt`:

```
  The dependency target "ArgumentParser" of target "CLICommands" does not
  exist.
```

kateinoigakukun: it's a totally different issue, the definition of `option(WASMKIT_BUILD_CLI "Build wasmkit-cli" ON)` should be before `if(WASMKIT_BUILD_CLI)` in the root `CMakeLists.txt`.

MaxDesiatov: Ugh, I can't say I love CMake 😅

PR #256, opened 2026-01-10
Tests: Migrate rest of test suites to use swift-testing

kateinoigakukun:


Tests/WITExtractorPluginTests/TestSupport.swift
```diff
@@ -52,10 +51,10 @@ struct TestSupport {
 
 func assertSwiftPackage(fixturePackage: String, _ trailingArguments: [String]) throws -> String {
     #if os(iOS) || os(watchOS) || os(tvOS) || os(visionOS)
-        throw XCTSkip("WITExtractor does not support platforms where Foundation.Process is unavailable")
+        fatalError("WITExtractor does not support platforms where Foundation.Process is unavailable")
     #else
         guard let config = TestSupport.Configuration.default else {
-            throw XCTSkip("Please create 'Tests/default.json'")
+            fatalError("Please create 'Tests/default.json'")
```
MaxDesiatov, line 57: Wouldn't we prefer to run rest of the tests by throwing as before if `Tests/default.json` is absent?

kateinoigakukun: This line won't be reachable because we skip them by `@Test(.disabled())`.

PR #266, opened 2026-01-16
WAT: basic Component Model definitions parsing

MaxDesiatov:
Currently this only includes component, core modules, core instances, and canon lifting/lowering definitions parsing. While the proposal test suite is added, it isn't wired up to WasmKit's test suite yet.

All component model code is hidden behind the new `ComponentModel` package trait, which is disabled by default.

A basic test is added that verifies recursive components text format parsing.

### DIFF
--- a/Sources/WAT/Parser/ComponentWatParser.swift
+++ b/Sources/WAT/Parser/ComponentWatParser.swift
@@ -1,6 +1,5 @@
 #if ComponentModel
 
-    import BasicContainers
     import ComponentModel
     import WasmParser
     import WasmTypes


Package.swift
```diff
@@ -65,7 +65,10 @@ let package = Package(
 
         .target(
             name: "WAT",
-            dependencies: ["WasmParser"],
+            dependencies: [
+                "ComponentModel",
```
MaxDesiatov:
```suggestion
                .target(name: "ComponentModel", condition: .when(traits: ["ComponentModel"])),
```

Package.swift
```diff
@@ -33,6 +33,7 @@ let package = Package(
     ],
     traits: [
         .default(enabledTraits: []),
+        "ComponentModel",
```
kateinoigakukun, line 36: I think we don't need a new trait because it doesn't bring any dependency. I'd prefer having a minimal build variations for maintainability.

MaxDesiatov: My concern here is that until implementation is finished, we have a risk of users misunderstanding the level of support for it. Additionally, the feature is still in Phase 1 per the proposals list https://github.com/WebAssembly/proposals

kateinoigakukun: It makes some sense, ok let's leave it as is.

PR #271, opened 2026-01-17
Fix all warnings

kateinoigakukun:
This PR fixes all compiler warnings in the codebase:

- Fixed unhandled file warnings by adding exclusions for LICENSE.txt and CMakeLists.txt files
- Updated plugin files to use URL instead of deprecated Path API
- Fixed deprecated String.init(validatingUTF8:) to use String.init(validatingCString:)
- Removed unreachable code in FileOperations.swift
- Fixed deprecated String.init(cString:) to use String(decoding:as: UTF8.self)
- Fixed deprecated assign(from:count:) to use update(from:count:) and removed unnecessary try
- Fixed unsafe pointer warning by using withUnsafeMutablePointer
- Fixed FilePath initialization to use platformString initializer
- Fixed dev_t conversion to preserve bit pattern for both signed and unsigned types

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -62,7 +62,7 @@ let package = Package(
         .testTarget(
             name: "WasmKitTests",
             dependencies: ["WasmKit", "WAT", "WasmKitFuzzing"],
-            exclude: ["ExtraSuite"]
+            exclude: ["ExtraSuite", "CMakeLists.txt"]
         ),
 
         .target(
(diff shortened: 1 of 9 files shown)


MaxDesiatov (approved):
Thanks!

Package.swift
MaxDesiatov, line 1: WDYT of enabling warnings as errors on most if not all modules here?

kateinoigakukun: I think we can enable it on CI but we should avoid always turning it on because it requires unsafeFlags usage.

PR #272, opened 2026-01-17
`mprotect`-based memory boundary checking

kateinoigakukun:
Use `mprotect`-based virtual memory for Wasm linear memory on macOS and Linux.

Instead of bounds-checking every load/store in software, reserve the full 4GB address space via `mmap(PROT_NONE)` and `mprotect` committed pages as the memory grows. Out-of-bounds accesses hit unmapped guard pages, triggering `SIGSEGV`/`SIGBUS` which a registered signal handler converts into a Wasm trap.

This adds `Unchecked` variants of all memory load/store instructions. The translator emits these when the target memory uses mprotect and the `memarg.offset` fits within the guard region, skipping the software bounds check entirely.

New `EngineConfiguration.memoryBoundsChecking` option (`.auto` / `.mprotect` / `.software`). Default is `.auto`, which uses mprotect on supported platforms and falls back to software checks otherwise. Memory64 always uses software checks.

## Signal Safety

`siglongjmp` from a signal handler does NOT run Swift destructors (`deinit`, `defer`, `withValue` closures) or ARC release operations on abandoned frames.

The direct-threaded path is safe: `wasmkit_tc_start` runs the dispatch loop in C trampolines, the `memory.withValue` lock is released before dispatch begins, and the unchecked instruction handlers (`memoryLoadUnchecked`/`memoryStoreUnchecked`) hold no resources. The only Swift scope on the stack is `withUnsafeMutablePointer(to: &self)` which merely pins a pointer.

The token-threaded path is problematic: `runTokenThreadedImpl` contains a Swift `while true { try doExecute(...) }` loop. A fault during an unchecked memory instruction causes `siglongjmp` to cut through Swift frames, which is undefined behavior in Swift. If the fault occurs mid-`withValue` on a `Mutex`-protected resource, the lock is never released.

For this reason, `Engine.init` will throw `EngineConfigurationError` when a user requests token-threading (including automatically selected one on platforms without direct threading) and explicit `mprotect`-based bounds checking.

### DIFF
--- a/Documentation/MprotectMemoryBoundsChecking.md
+++ b/Documentation/MprotectMemoryBoundsChecking.md
@@ -1,116 +1,87 @@
-# mprotect-based memory boundary checking
+# mprotect-based memory bounds checking
 
 ## Background
 
-WasmKit supports two strategies for linear-memory bounds checking. The default software path performs an explicit bounds check for every load/store in `Execution.memoryLoad` / `Execution.memoryStore` (`Sources/WasmKit/Execution/Instructions/Memory.swift`).
+WasmKit has two linear-memory allocation strategies:
 
-On platforms where virtual memory protection and synchronous fault handling are available, we can remove these per-access checks by:
+- `software`: linear memory accesses are checked explicitly in the interpreter before touching memory.
+- `mprotect`: wasm32 linear memories are backed by a large virtual address reservation whose inaccessible tail can be used to turn out-of-bounds accesses into synchronous faults.
 
-1. Reserving a large virtual address range for each linear memory.
-2. Committing (read/write) only the currently-accessible portion.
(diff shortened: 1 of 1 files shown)


Documentation/MprotectMemoryBoundsChecking.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Platform support matrix
+
+Planned initial support:
+
+- macOS (arm64/x86_64): enabled
+- Linux (arm64/x86_64): enabled
+
+Disabled (fallback to software checks):
+
```
MaxDesiatov, line 69:
```suggestion

- Windows (requires a separate implementation for WinAPI)
```

PR #273, opened 2026-01-18
Add `wat2wasm` example compatible with Embedded Swift

MaxDesiatov:
This is an example which demonstrates use of some (but not all) modules of WasmKit in Embedded Swift context.


Sources/WAT/WatParserError.swift
```diff
@@ -1,6 +1,6 @@
 /// An error type thrown during WAT (WebAssembly Text Format) parsing.
 public struct WatParserError: Swift.Error {
-    package let message: String
+    public let message: String
```
MaxDesiatov:
```suggestion
    package let message: String
```

kateinoigakukun (approved):

PR #276, opened 2026-01-18
Remove generic `Stream` and `StreamError` types

MaxDesiatov:
Both `Stream` and `StreamError` are over-generalization of generic streams, while only `ByteStream` is actually used in `WasmParser`, and `StreamError` is always used with `UInt8` type parameter.

Let's clean these up as unused, `WasmParser` is not in the business of providing general stream APIs.

`StreamError` is now merged with `WasmParserError`, which simplifies future adoption of typed throws to unblock https://github.com/swiftwasm/WasmKit/pull/273.

While we're at it, `ByteStream` now conforms to `~Copyable`, but ideally we'd want it to return `Span<UInt8>` and not `ArraySlice` from its methods. Bumping required Swift version to 6.2 is required for that, which we're not ready to do just yet, so `~Escapable` constraint is not added in this PR.

### DIFF
--- a/Sources/WasmParser/WasmParser.swift
+++ b/Sources/WasmParser/WasmParser.swift
@@ -208,7 +208,8 @@ public struct WasmFeatureSet: OptionSet, Sendable {
 
 /// An error that occurs during parsing of a WebAssembly binary
 public struct WasmParserError: Swift.Error {
-    public struct Message: Sendable {
+    @usableFromInline
+    struct Message: Sendable {
         let text: String
 
         init(_ text: String) {


Sources/WasmParser/WasmParser.swift
```diff
@@ -207,29 +207,55 @@ public struct WasmFeatureSet: OptionSet, Sendable {
 }
 
 /// An error that occurs during parsing of a WebAssembly binary
-public struct WasmParserError: Swift.Error {
```
kateinoigakukun, line 210: I'd still want to avoid exposing public `enum` here to minimize public interface. Also I think those new error kinds should have binary offset as well.

Sources/WasmParser/WasmParser.swift
```diff
@@ -208,28 +208,70 @@ public struct WasmFeatureSet: OptionSet, Sendable {
 
 /// An error that occurs during parsing of a WebAssembly binary
 public struct WasmParserError: Swift.Error {
-    @usableFromInline
-    struct Message: Sendable {
+    public struct Message: Sendable {
```
kateinoigakukun:
```suggestion
    struct Message: Sendable {
```

Sources/WasmParser/WasmParser.swift
```diff
@@ -208,28 +208,70 @@ public struct WasmFeatureSet: OptionSet, Sendable {
...
+            return result
+        case .unexpectedByte(let byte, let index, let expected):
+            var result = "Unexpected byte \(byte.hexString) at index \(index.hexString)."
+            if let expected, expected.count > 0 {
+                result.append(contentsOf: " Expected one of \(expected.map {$0.hexString})")
+            }
+            return result
+        case .unclassified(let error):
+            return "\(error)"
+        }
```
kateinoigakukun, line 277: Non-message errors should print `at offset ...` when available

PR #277, opened 2026-01-18
Use unified `WasmKitError` for typed throws

MaxDesiatov:
The main reason for using single array type is that combining error types from different visitor types is impractical in certain functions that combine those visitors, and probably also detrimental for code size and performance.

Fortunately, existing error types fit well within the unified type, so mostly it's a text replacement to use `WasmKitError` everywhere.

Unblocks https://github.com/swiftwasm/WasmKit/pull/295 and https://github.com/swiftwasm/WasmKit/pull/273.


Sources/WasmParser/InstructionVisitor.swift
```diff
@@ -548,373 +548,372 @@ public enum Instruction: Equatable {
...
+    public mutating func visitSimdLane(_ simdLane: Instruction.SimdLane, lane: UInt8) throws(WasmKitError) { return try self.visit(.simdLane(simdLane, lane: lane)) }
+    public mutating func visitSimdMemLane(_ simdMemLane: Instruction.SimdMemLane, memarg: MemArg, lane: UInt8) throws(WasmKitError) { return try self.visit(.simdMemLane(simdMemLane, memarg: memarg, lane: lane)) }
 }
 
 /// A visitor for WebAssembly instructions.
 ///
 /// The visitor pattern is used while parsing WebAssembly expressions to allow for easy extensibility.
 /// See the expression parsing method ``Code/parseExpression(visitor:)``
 public protocol InstructionVisitor: ~Copyable {
-    associatedtype VisitorError: Error
```
kateinoigakukun, line 677: Hmm, I still think we should not enforce the use of the unified Error type here because the visitor is meant to be consumed by WasmParser module users as well.

kateinoigakukun (commented):
I still not get why it's not practical to combine visitors 🤔 Do you have any example?

MaxDesiatov:
> I still not get why it's not practical to combine visitors 🤔 Do you have any example?

Consider this in `BinaryInstructionDecoder.swift:216`

```swift
func parseBinaryInstruction(
    visitor: inout some InstructionVisitor & ~Copyable,
    decoder: inout some BinaryInstructionDecoder
) throws(WasmKitError) -> Bool {
    visitor.binaryOffset = decoder.offset
    let opcode0 = try decoder.claimNextByte()
    switch opcode0 {
    case 0x00:
        try visitor.visitUnreachable()
    case 0x01:
        try visitor.visitNop()
    case 0x02:
        let (blockType) = try decoder.visitBlock()
        try visitor.visitBlock(blockType: blockType)
    case 0x03:
        let (blockType) = try decoder.visitLoop()
        try visitor.visitLoop(blockType: blockType)
```

We don't have concrete visitor and decoder types here, nor we have their concrete error types. Even if we introduce `EitherError` for combining them, it becomes quite tedious to handle that combination in function's body.

Sources/WasmParser/InstructionVisitor.swift
MaxDesiatov: See `parseBinaryInstruction` example below that uses this type and has trouble with expressing combined error type via this associated type. IMO users of `WasmParser` should be ok with existing `WasmKitError`, and we can expand its API if needed (which is already general enough).

Sources/WasmTypes/Location.swift
kateinoigakukun, line 1: Do we still need to move this? I'd like to keep the WasmTypes minimal

Sources/WasmParser/WasmParser.swift
```diff
@@ -157,18 +112,41 @@ public struct ExpressionParser {
...
-        isLastEnd = try parser.parseInstruction(visitor: &visitor)
-        let shouldContinue = try !parser.stream.hasReachedEnd()
-        if !shouldContinue {
-            guard isLastEnd == true else {
-                throw WasmParserError(.endOpcodeExpected, offset: offset)
-            }
+    public mutating func parse() throws(WasmParserError) -> Visit? {
+        if reachedEnd { return nil }
+        let instructionOffset = offset
+        let instruction = try parser.parseInstruction()
```
kateinoigakukun, line 123: Can you take a profile just in case? Holding a `Instruction` and dispatch visit based on the data adds extra branches at runtime. Perhaps we need to return a closure from `parseBinaryInstruction` that captures parsed operand info, takes a visitor, and dispatches a corresponding visit method

Sources/WAT/WAT.swift
```diff
@@ -1,4 +1,5 @@
 import WasmParser
+import WasmTypes
```
kateinoigakukun: We don't need this any more?

Sources/WAT/Lexer.swift
```diff
@@ -1,4 +1,5 @@
 import WasmParser
+import WasmTypes
```
kateinoigakukun: ditto

Sources/WasmKit/Translator.swift
```diff
@@ -1237,35 +1238,29 @@ struct InstructionTranslator: InstructionVisitor {
...
             emit(.onEnter(functionIndex))
         }
         var parser = ExpressionParser(code: code)
-        var offset = parser.offset
-        do {
-            while try parser.visit(visitor: &self) {
-                offset = parser.offset
-            }
-        } catch var error as ValidationError {
-            error.offset = offset
```
kateinoigakukun, line 1254: We should not discard binary offset

MaxDesiatov: I've updated it to preserve the offset

Sources/WasmParser/WasmParser.swift
MaxDesiatov: No significant performance difference with CoreMark, would you like a more thorough analysis?

MaxDesiatov:
@kateinoigakukun gentle ping 🙂

Sources/WAT/WAT.swift
kateinoigakukun: @MaxDesiatov Seems still remaining?

MaxDesiatov:
```suggestion
```

Sources/WAT/Lexer.swift
MaxDesiatov:
```suggestion
```

Sources/WAT/WAT.swift
MaxDesiatov: Thanks, addressed

Sources/WAT/Lexer.swift
MaxDesiatov: Addressed as well

Sources/WasmTypes/Location.swift
kateinoigakukun: @MaxDesiatov WDYT?

MaxDesiatov: I think this has been addressed, but current GitHub UI no longer shows comments on files as outdated

kateinoigakukun: Seems like the file is still in WasmTypes? `Sources/WasmTypes/Location.swift`

MaxDesiatov: Which commit are you comparing against? On c90ea68394a6e89cdb539cc28ba945d8319916b5 I see this:

```
> fd Location
Sources/WAT/Location.swift
```

kateinoigakukun: Oh, I cannot see that anymore. Maybe I didn't reload? Sorry :bow:

kateinoigakukun (approved):

PR #278, opened 2026-01-18
Add `WasmParserBenchmark`, CI job to verify that benchmarks build

MaxDesiatov:
This benchmark focuses on `WasmParser` performance, extracting a diverse set of Wasm modules to test on from the spectest suite.

Other benchmarks in `Benchmarks/Package.swift` were updated for the new Swift version and platform requirements, together with changes necessary to adopt the new `WASIBridgeToHost` API.

Also added `build-benchmark` CI job that ensures that benchmarks code no longer bit-rots.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/checkout@v4
       - id: setup-development
         run: |
-          brew install wabt
+          brew install wabt jemalloc
           curl -O https://download.swift.org/swiftly/darwin/swiftly.pkg && \
             installer -pkg swiftly.pkg -target CurrentUserHomeDirectory
 
@@ -224,7 +224,10 @@ jobs:
       - run: ./build-exec swift test ${{ matrix.test-args }}
       
       - if: matrix.build-benchmarks
-        run: ./build-exec swift package --package-path Benchmarks benchmark
+        run: |
+          ./build-exec apt-get update
+          ./build-exec apt-get install -y libjemalloc-dev
+          ./build-exec swift package --package-path Benchmarks benchmark
 
       - if: matrix.build-dev-dashboard
         run: ./build-exec ./CI/build-dev-dashboard.sh


Benchmarks/Benchmarks/WasmParserBenchmark/WasmParserBenchmark.swift
```diff
@@ -0,0 +1,67 @@
+// Benchmark boilerplate generated by Benchmark
+
```
MaxDesiatov:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -337,3 +337,14 @@ jobs:
...
       - name: Build with the Swift SDK
         run: swift build --swift-sdk "$(swiftc -print-target-info | jq -r '.swiftCompilerTag')_wasm" --product wasmkit-cli
+
+  build-benchmarks:
+    runs-on: ubuntu-24.04
+    container:
+      image: swift:6.2-noble
+    steps:
+      - uses: actions/checkout@v4
+      - run: apt-get update && apt-get install -y python3
```
MaxDesiatov:
```suggestion
      - run: apt-get update && apt-get install -y python3 libjemalloc-dev
```

.github/workflows/main.yml
```diff
@@ -337,3 +337,14 @@ jobs:
         run: swift sdk install https://download.swift.org/swift-6.2.3-release/wasm-sdk/swift-6.2.3-RELEASE/swift-6.2.3-RELEASE_wasm.artifactbundle.tar.gz --checksum 394040ecd5260e68bb02f6c20aeede733b9b90702c2204e178f3e42413edad2a
       - name: Build with the Swift SDK
         run: swift build --swift-sdk "$(swiftc -print-target-info | jq -r '.swiftCompilerTag')_wasm" --product wasmkit-cli
+
+  build-benchmarks:
```
kateinoigakukun: I'm hesitant to add more CI job because we are already running out most of the runners on our free org plan, and we can't run CI for multiple PRs simultaneously even now. We need to optimize CI usage before adding more.

Benchmarks/Package.swift
```diff
@@ -1,13 +1,15 @@
...
 
 let package = Package(
     name: "Benchmarks",
-    platforms: [.macOS(.v13)],
+    platforms: [.macOS(.v14)],
     dependencies: [
         .package(name: "WasmKit", path: "../"),
         .package(url: "https://github.com/ordo-one/package-benchmark", .upToNextMajor(from: "1.4.0")),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.92.2"),
+        .package(url: "https://github.com/apple/swift-system.git", from: "1.6.4"),
```
kateinoigakukun: I'd like to keep the benchmark suite with minimal deps unless there are strong necessities so that we can run benchmark iteratively with minimal build cost.

.github/workflows/main.yml
MaxDesiatov: What's the alternative then to keep these benchmarks up to date? With active development, bit-rot is easy to introduce. Bumping Swift version and recent WASI in-memory FS changes did so. Should we move this build step to some of the existing jobs?

kateinoigakukun: Let's move it to one of existing jobs

PR #284, opened 2026-01-26
Include threads proposal tests in runtime spectest

kateinoigakukun:


Package.swift
```diff
@@ -20,7 +20,7 @@ let cliCommandsTarget = Target.target(
 
 let package = Package(
     name: "WasmKit",
-    platforms: [.macOS(.v14), .iOS(.v13)],
+    platforms: [.macOS(.v15), .iOS(.v18)],
```
MaxDesiatov, line 23: Do you think it's viable to annotate each function with availability attributes? IIUC it's not just that but also stored `Mutex` properties which makes types using them not really back-deployable.

kateinoigakukun: Can you check what's the minimum supported Apple-platform version of other toolchain binaries?

MaxDesiatov: SwiftPM right this moment specifies `[.macOS(.v14), .iOS(.v17), .macCatalyst(.v17),]` and IIUC more or less follows current-2 pattern https://github.com/swiftlang/swift-package-manager/commit/59f3e8d8a3fdc102eb2c3af4da46a4cde54b9cb1.

At the same time I don't think it's tenable to support Wasm threads while reinventing our own platform-independent `Mutex`. I propose bumping to macOS 15 now in `main` and to tag some future 0.3.0 release from it that would follow SwiftPM macOS bump. We merge this PR now (if you're fine with it), and if we need to tag a release urgently before that for macOS 14, we tag it off a new `release/0.2` branch created from existing 0.2.0 release with cherry-picks.

MaxDesiatov: Also, having threads in `main` influences LLVM backend design quite significantly. Keeping this PR open and maintaining both in parallel would also be problematic.

PR #287, opened 2026-02-03
Add `Vendor/component-model` test suite dependencies

MaxDesiatov:
As WABT doesn't support Wasm Component Model, to run its test-suite we'll need to use `wasm-tool` in addition to WABT. Fortunately, prebuilt Wasm binaries are available, so we don't have to build it and it's also virtualized thanks to WASI.

`Vendor/checkout-dependency` is modified to support downloading and extracting tarball dependencies

### DIFF
--- a/Vendor/checkout-dependency
+++ b/Vendor/checkout-dependency
@@ -20,8 +20,6 @@ class CheckoutDependency:
 
     def verify_sha256(self, file_path, expected_sha256):
         """Verify SHA256 checksum of a file"""
-        if expected_sha256.startswith("sha256:"):
-            expected_sha256 = expected_sha256[7:]
 
         sha256_hash = hashlib.sha256()
         with open(file_path, "rb") as f:


Vendor/dependencies.json
```diff
@@ -14,6 +14,16 @@
...
     "categories": ["default"]
   },
+  "component-model": {
+    "repository": "https://github.com/WebAssembly/component-model.git",
+    "revision": "0b951b72b3f4019edf386a93585782587d221a0c",
+    "categories": ["component-model"]
+  },
+  "wasm-tools": {
+    "url": "https://github.com/bytecodealliance/wasm-tools/releases/download/v1.244.0/wasm-tools-1.244.0-wasm32-wasip1.tar.gz",
+    "sha256": "sha256:010f8a3c591f3070d06a1c13830c8a61dcbf6527a62ab289970ed3341f0008e7",
```
kateinoigakukun:
```suggestion
    "sha256": "010f8a3c591f3070d06a1c13830c8a61dcbf6527a62ab289970ed3341f0008e7",
```

`sha256:` prefix feels like redundant?

Vendor/checkout-dependency
```diff
@@ -9,22 +9,69 @@ import os
...
     def __init__(self, dependencies_file):
         self.dependencies_file = dependencies_file
         self.dependencies = json.load(open(dependencies_file))
 
-    def checkout(self, dependency_name):
-        dependency = self.dependencies.get(dependency_name)
+    def verify_sha256(self, file_path, expected_sha256):
+        """Verify SHA256 checksum of a file"""
+        if expected_sha256.startswith("sha256:"):
+            expected_sha256 = expected_sha256[7:]
```
MaxDesiatov:
```suggestion
```

PR #288, opened 2026-02-03
Re-enable all tests for nightly Swift toolchains

MaxDesiatov:
Removed conditional check for `const.wast` in Spectest.swift


kateinoigakukun (approved):
As long as CI is green

PR #290, opened 2026-02-06
Add component binary encoder, enable some CM spec tests 

MaxDesiatov:
The main complexity is caused by the fact that CM binaries and normalized CM WAT require significant amount of normalization when compared to source WAT: inline exports/imports, aliases etc. Unlike core modules, CM binaries don't have sections consolidated in increasing ID order, but can be disjoint and interleaved. At the same time, some definitions are still batched when they're consecutive.

With this implementation we can enable a couple of tests from the spec (https://github.com/WebAssembly/component-model/tree/main/test/wasm-tools) and can verify that WasmKit is compatible on binary level with existing implementations.

New `WasmTools` module is added to wrap some of these tools to be called from tests or when debugging locally. This tool is only linked into tests when `ComponentModel` trait is enabled. An alternative could be to fold it back into the test module, but that makes it more complex to create of ad-hoc analysis/debugging tools that reuse vendored `wasm-tools.wasm` binary. This binary doesn't require building from source and is also helpfully virtualized, so everything happens in-memory and no FS cleanup is needed after running it.

### DIFF
--- a/Sources/WAT/BinaryEncoding/ComponentEncoder.swift
+++ b/Sources/WAT/BinaryEncoding/ComponentEncoder.swift
@@ -793,8 +793,8 @@
                     // 1. Function imports (at indices 0, 1, ...)
                     // 2. Function aliases (continuing indices)
                     // Since we emit aliases before canon.lower, the aliasIndexInArray
-                    // is the correct index (offset by any function imports, but we
-                    // don't have function imports in these test cases yet)
+                    // is the correct index (offset by any function imports)
+                    #warning("No function imports in enabled test cases yet to exercise this code path fully")
                     encoder.writeUnsignedLEB128(UInt32(aliasIndexInArray))
 
                     try encoder.encodeVector(canonDef.options) { option, encoder throws(WatParserError) in


Sources/ComponentModel/ComponentType.swift
```diff
@@ -97,26 +97,94 @@ public enum ComponentType: Hashable {
     case variant([ComponentCaseField])
     case resource(destructor: ComponentFuncIndex)
 
-    indirect case `func`(ComponentFuncType)
-
     case indexed(ComponentTypeIndex)
+
+    /// Returns true if this is a primitive type that can be inlined in valtypes
```
MaxDesiatov:
```suggestion
    /// Returns `true` if this is a primitive type that can be inlined in valtypes during binary encoding
```

Sources/ComponentModel/ComponentType.swift
```diff
@@ -97,26 +97,94 @@ public enum ComponentType: Hashable {
...
+    case u8(UInt8)
+    case u16(UInt16)
+    case u32(UInt32)
+    case u64(UInt64)
+    case float32(Float)
+    case float64(Double)
+    case char(Unicode.Scalar)
+    case string(String)
+
+    // Composite types
```
MaxDesiatov, line 177:
```suggestion
    // Composite types
    // TODO: store strings in an interner, create a resizable bitset type for `case flags`
```

Sources/WAT/BinaryEncoding/ComponentEncoder.swift
```diff
@@ -0,0 +1,1471 @@
+#if ComponentModel
+    import ComponentModel
+    import WasmTypes
+
+    public struct ComponentEncoder {
```
MaxDesiatov, line 6:
```suggestion
    /// Binary component encoder, implementing CM proposal `Binary.md` spec: https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md
    public struct ComponentEncoder {
```

Sources/WAT/BinaryEncoding/ComponentEncoder.swift
```diff
@@ -0,0 +1,1471 @@
...
+            writeHeader()
+
+            // Collect metadata needed for encoding
+            let fields = try groupFields(component.fields)
+
+            // Build core instance index mapping (parser index -> binary index)
+            // This accounts for inline export instances that shift indices
+            let coreInstanceIndexMapping = buildCoreInstanceIndexMapping(component: component, fields: fields)
+
+            let coreFuncAliases = try collectCoreFuncAliases(from: fields.canons, component: component, coreInstanceIndexMapping: coreInstanceIndexMapping)
```
MaxDesiatov:
```suggestion
            let coreFuncAliases = try collectCoreFuncAliases(
                from: fields.canons,
                component: component,
                coreInstanceIndexMapping: coreInstanceIndexMapping
            )
```

Sources/WAT/BinaryEncoding/ComponentEncoder.swift
```diff
@@ -0,0 +1,1471 @@
...
+                    encoder.output.append(0x01)  // canon.lower
+                    encoder.output.append(0x00)  // sub-opcode
+
+                    // The function index refers to the component functions index space.
+                    // Component functions come from:
+                    // 1. Function imports (at indices 0, 1, ...)
+                    // 2. Function aliases (continuing indices)
+                    // Since we emit aliases before canon.lower, the aliasIndexInArray
+                    // is the correct index (offset by any function imports, but we
+                    // don't have function imports in these test cases yet)
```
MaxDesiatov:
```suggestion
                    // is the correct index (offset by any function imports)
                    #warning("No function imports in enabled test cases yet to exercise this code path fully")
```

PR #291, opened 2026-02-09
Add component binary parser and parsed representation

MaxDesiatov:
To enable more tests, we need to be able to load a component. While we could implement a loader that uses WAT parser output, the most common use case is binary component loading, not loading from WAT. This also follows our existing flow for loading core modules from WAT: core module WAT -> core module binary -> `Engine`/`Store`/`Instance`.

The loader and relevant types will be added in a future PR, while this one is focused on adding `ComponentParser` type available when `ComponentModel` trait is enabled.

Also cleaned up `CoreAliasSort` type, which overlaps with `CoreDefSort`, and we should reuse the latter where possible.

### DIFF
--- a/Sources/WasmParser/ComponentParser.swift
+++ b/Sources/WasmParser/ComponentParser.swift
@@ -56,7 +56,7 @@
         case customSection(CustomSection)
 
         /// An embedded core module (section 1)
-        case coreModule([UInt8])
+        case coreModule(ArraySlice<UInt8>)
 
         /// Core instance definitions (section 2)
         case coreInstanceSection([CoreInstanceDefinition])
@@ -65,7 +65,7 @@
         case coreTypeSection([CoreTypeDef])
 
         /// A nested component (section 4)
-        case component([UInt8])
+        case component(ArraySlice<UInt8>)
 
         /// Component instance definitions (section 5)
         case instanceSection([ComponentInstanceDefinition])
@@ -954,7 +954,7 @@
 
                 case 1:  // core module
                     let moduleBytes = try stream.consume(count: Int(sectionSize))
-                    payload = .coreModule(Array(moduleBytes))
+                    payload = .coreModule(moduleBytes)
 
                 case 2:  // core instance
                     payload = .coreInstanceSection(try parseCoreInstanceSection())
@@ -965,7 +965,7 @@
 
                 case 4:  // component
                     let componentBytes = try stream.consume(count: Int(sectionSize))
-                    payload = .component(Array(componentBytes))
+                    payload = .component(componentBytes)
 
                 case 5:  // instance
                     payload = .instanceSection(try parseInstanceSection())


Sources/WasmParser/ComponentParser.swift
```diff
@@ -0,0 +1,1005 @@
...
+
+            case 0x66:  // stream
+                let elemIdx = try parseOptionalValType()
+                return .stream(element: elemIdx, end: nil)
+
+            case 0x65:  // future
+                let elemIdx = try parseOptionalValType()
+                return .future(elemIdx)
+
+            // Primitive types (fall through)
```
MaxDesiatov:
```suggestion
            // Primitive types
```

Sources/WasmParser/ComponentParser.swift
```diff
@@ -0,0 +1,1005 @@
+#if ComponentModel
+    import WasmTypes
+    import ComponentModel
+
+    /// A streaming parser for WebAssembly Component Model binary format.
```
MaxDesiatov, line 5:
```suggestion
    /// A streaming parser for WebAssembly Component Model binary format.
    /// For proposed production rules refer to https://github.com/WebAssembly/component-model/blob/7479890cb506d0f8f687595a4d41361ff8a2a194/design/mvp/Binary.md
```

Sources/WasmParser/ComponentParser.swift
```diff
@@ -0,0 +1,1005 @@
...
+                case .scalarValue(let scalar): name.append(Character(scalar))
+                case .emptyInput: break loop
+                case .error: throw makeError(.invalidUTF8(Array(nameBytes)))
+                }
+            }
+
+            // Handle version suffix if present
+            if prefix == 0x01 {
+                let suffixLen: UInt32 = try parseUnsigned()
+                _ = try stream.consume(count: Int(suffixLen))  // Skip version suffix for now
```
MaxDesiatov, line 485:
```suggestion
                #warning("Version suffix skipped in `parseImportExportName` as not implemented yet")
                _ = try stream.consume(count: Int(suffixLen))  // Skip version suffix for now
```

Sources/WasmParser/WasmParser.swift
```diff
@@ -1430,3 +1430,60 @@ public struct NameSectionParser<Stream: ByteStream> {
...
+///
+/// - Parameter filePath: Path to the WebAssembly binary file
+/// - Returns: The detected file type
+/// - Throws: If the file cannot be opened or read
+public func detectWasmFileType(filePath: FilePath) throws -> WasmFileType {
+    let fileHandle = try FileDescriptor.open(filePath, .readOnly)
+    defer { try? fileHandle.close() }
+
+    // Use a tuple to avoid heap allocation - 8 bytes on stack
+    var header: (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8) = (0, 0, 0, 0, 0, 0, 0, 0)
```
MaxDesiatov, line 1461:
```suggestion
    // TODO: needs a `SmallArray` abstraction until `InlineArray` becomes available after dropping support for macOS 15.
    var header: (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8) = (0, 0, 0, 0, 0, 0, 0, 0)
```

Sources/WasmParser/ComponentParser.swift
```diff
@@ -0,0 +1,1007 @@
...
+                let sectionStart = stream.currentIndex
+
+                let payload: ComponentParsingPayload
+                switch sectionID {
+                case 0:  // custom
+                    payload = .customSection(try parseCustomSection(size: sectionSize))
+
+                case 1:  // core module
+                    let moduleBytes = try stream.consume(count: Int(sectionSize))
+                    payload = .coreModule(Array(moduleBytes))
```
kateinoigakukun: Let's avoid copying the entire module bytes.

Sources/WasmParser/ComponentParser.swift
```diff
@@ -0,0 +1,1007 @@
...
+                case 2:  // core instance
+                    payload = .coreInstanceSection(try parseCoreInstanceSection())
+
+                case 3:  // core type
+                    let coreTypes = try parseVector { try parseCoreTypeDef() }
+                    payload = .coreTypeSection(coreTypes)
+
+                case 4:  // component
+                    let componentBytes = try stream.consume(count: Int(sectionSize))
+                    payload = .component(Array(componentBytes))
```
kateinoigakukun: ditto

Sources/WasmParser/ComponentParser.swift
MaxDesiatov: Nice catch! `ArraySlice` good enough?

PR #296, opened 2026-02-11
Add `ComponentInstance`/`ComponentLoader` types

MaxDesiatov:
This allows loading components from binary representation and calling component functions taking/returning component values. This is going to unblock multiple spec tests enabled in future PRs.

### DIFF
--- a/Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
+++ b/Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
@@ -871,19 +871,19 @@
         }
 
         fileprivate func liftUTF16LE(pointer: UInt32, byteCount: Int, codeUnitCount: Int) throws -> String {
-                // Read UTF-16LE code units from memory
-                var codeUnits: [UInt16] = []
-                codeUnits.reserveCapacity(codeUnitCount)
+            // Read UTF-16LE code units from memory
+            var codeUnits: [UInt16] = []
+            codeUnits.reserveCapacity(codeUnitCount)
 
-                self.withUnsafeBufferPointer(offset: UInt(pointer), count: byteCount) { buffer in
-                    for i in 0..<codeUnitCount {
-                        let offset = i * 2
-                        let low = UInt16(buffer[offset])
-                        let high = UInt16(buffer[offset + 1])
-                        let codeUnit = low | (high << 8)  // Little-endian
-                        codeUnits.append(codeUnit)
-                    }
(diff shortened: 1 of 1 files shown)


Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
```diff
@@ -0,0 +1,1446 @@
+// Canonical ABI lifting/lowering implementation. See `CanonicalABI.md` for more details:
+// https://github.com/WebAssembly/component-model/blob/a565206aea8190e13d3a297e138261455a8b80ad/design/mvp/CanonicalABI.md
+
+#if ComponentModel
+    #if canImport(Foundation)
+        import Foundation
```
kateinoigakukun: Why do we need `Foundation` here?

Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
kateinoigakukun, line 1: Would you mind considering using  https://github.com/swiftwasm/WasmKit/blob/18fa60bba7c074047639c0ba12e6a0187d1521b6/Sources/WIT/CanonicalABI/CanonicalLifting.swift to centralize the compositional type handling?

Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
MaxDesiatov: It enables UTF-16 lifting, there's a similar `#if canImport(Foundation)` in functions using it, otherwise lifting will throw

kateinoigakukun: Can we use `String(decoding: scalars, as: UTF16.self)` instead?

Sources/WasmKit/Execution/ComponentModel/Lift+Lower.swift
MaxDesiatov: That `protocol CanonicalLifting` is tightly coupled to `WITType`, which is itself tied to WIT type resolution. At the same time `ComponentValueType` is coupled to `ComponentTypeIndex` and resolution via indices used in runtime's component representation.

IIUC WIT is a compile-time concept, and we don't want the loader, the runtime, or anything in `WasmKit` module to operate on `WITType` instead of `ComponentValueType`.

That leaves migration of `CanonicalLifting` to `ComponentValueType` as the other option (other than merging this PR as is or with minor changes), but also quite disruptive. Do you have strong preference for that refactoring to happen before we have a working component loader?

PR #300, opened 2026-02-12
`.github/workflows/cflite_pr.yml`: use older runner version

MaxDesiatov:
Downgraded os runner image to be compatible with `Dockerfile` used for CFLite fuzzer.


MaxDesiatov:
> down-versioning ubuntu-latest to ubuntu-22.04 might be a workaround?

Seems like that worked, thanks!

PR #301, opened 2026-02-15
Add component WAT support to wat2wasm CLI

MaxDesiatov:
Extends the `wat2wasm` function and CLI command to support both core modules and CM components. Uses parser-based detection: tries module parsing first, then falls back to component parsing if that fails.

### DIFF
--- a/Sources/WAT/WAT.swift
+++ b/Sources/WAT/WAT.swift
@@ -40,21 +40,30 @@ public func wat2wasm(
     features: WasmFeatureSet = .default,
     options: EncodeOptions = .default
 ) throws -> [UInt8] {
-    do {
-        var wat = try parseWAT(input, features: features)
-        return try encode(module: &wat, options: options)
-    } catch {
-        #if ComponentModel
-            // Module parsing failed, try as component
+    #if ComponentModel
+        // Look ahead to determine if this is a component or module
+        var peekParser = Parser(input)
+        let isComponent: Bool
+        if try peekParser.takeParenBlockStart("component") {
+            isComponent = true
+        } else {
+            isComponent = false
+        }
+
+        if isComponent {
             var parser = Parser(input)
             let componentParser = ComponentWatParser(features: features)
             let componentDef = try componentParser.parse(&parser)
             var encoder = ComponentEncoder()
             return try encoder.encode(componentDef, options: options)
-        #else
-            throw error
-        #endif
-    }
+        } else {
+            var wat = try parseWAT(input, features: features)
+            return try encode(module: &wat, options: options)
+        }
+    #else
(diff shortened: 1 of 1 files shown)


Sources/WAT/WAT.swift
```diff
@@ -15,6 +15,10 @@ public struct EncodeOptions: Sendable {
 }
 
 /// Transforms a WebAssembly text format (WAT) string into a WebAssembly binary format byte array.
+///
+/// This function supports both core modules and Component Model components (when the ComponentModel
```
MaxDesiatov:
```suggestion
/// This function supports both core modules and Component Model components (when the `ComponentModel`
```

Sources/CLICommands/Wat2wasm.swift
```diff
@@ -58,7 +69,10 @@ package struct Wat2wasm: ParsableCommand {
     )
     var output: String?
 
-    @Flag(help: "Include the name section in the output binary.")
+    @Flag(
+        inversion: .prefixedNo,
```
MaxDesiatov:
```suggestion
        inversion: .prefixedEnableDisable,
```

Sources/WAT/WAT.swift
```diff
@@ -36,8 +40,21 @@ public func wat2wasm(
...
+        var wat = try parseWAT(input, features: features)
+        return try encode(module: &wat, options: options)
+    } catch {
+        #if ComponentModel
+            // Module parsing failed, try as component
+            var parser = Parser(input)
+            let componentParser = ComponentWatParser(features: features)
+            let componentDef = try componentParser.parse(&parser)
+            var encoder = ComponentEncoder()
+            return try encoder.encode(componentDef, options: .init())
```
MaxDesiatov:
```suggestion
            return try encoder.encode(componentDef, options: options)
```

Sources/WAT/WAT.swift
```diff
@@ -36,8 +40,21 @@ public func wat2wasm(
...
     options: EncodeOptions = .default
 ) throws -> [UInt8] {
-    var wat = try parseWAT(input, features: features)
-    return try encode(module: &wat, options: options)
+    do {
+        var wat = try parseWAT(input, features: features)
+        return try encode(module: &wat, options: options)
+    } catch {
+        #if ComponentModel
+            // Module parsing failed, try as component
```
kateinoigakukun: Ideally, we should look ahead a few tokens instead of relying on parsing failure
https://github.com/bytecodealliance/wasm-tools/blob/76927bf4bdbddf4b15f835c5eddfffbdfe3bdbd5/crates/wast/src/wat.rs#L41

PR #305, opened 2026-02-22
Add `wasm2wat` function and CLI

MaxDesiatov:
Counterpart to the existing [`wat2wasm`](https://swiftpackageindex.com/swiftwasm/wasmkit/main/documentation/wat/wat2wasm(_:features:options:)) function.


Sources/WAT/Printer/ModuleCollector.swift
```diff
@@ -0,0 +1,100 @@
+import WasmParser
+import WasmTypes
+
+/// Collected information from all sections of a WebAssembly binary module.
+struct ModuleInfo {
```
MaxDesiatov: @kateinoigakukun do you have a strong opinion whether introducing this new `internal` struct is fine, or should we collect into the existing `public struct Wat` that uses `NameMapping`?

MaxDesiatov: Reviewing it more, I see that `Wat` has `let parser: Parser` text parser, which doesn't make sense to have when converting in an opposite direction: from binary to text. I don't think we can reuse `Wat` here.

Sources/WAT/Printer/TextInstructionVisitor.swift
```diff
@@ -0,0 +1,931 @@
...
+            emit(tbl == 0 ? "table.get" : "table.get \(tbl)")
+        case .tableSet(let tbl):
+            emit(tbl == 0 ? "table.set" : "table.set \(tbl)")
+        case .tableGrow(let tbl):
+            emit(tbl == 0 ? "table.grow" : "table.grow \(tbl)")
+        case .tableSize(let tbl):
+            emit(tbl == 0 ? "table.size" : "table.size \(tbl)")
+
+        // ── Atomics ──────────────────────────────────────────────────────────
+        case .atomicFence:
```
MaxDesiatov, line 208: @kateinoigakukun with some refactoring a lot of these instructions could be migrated to `wasmgen` in a future PR

MaxDesiatov: Looking at `wasmgen` more, it seems like there is a lot of special handling required here that `wasmgen` can't easily provide. We'll have to add more fields to `Instructions.json`, not sure that's worth it for something like `wasm2wat`, which is less load-bearing and more of a debugging utility.

Sources/WAT/Printer/ModuleCollector.swift
kateinoigakukun: Do we need to collect all information here before emitting text? Can we streamingly emit text without collecting them?

Sources/WAT/Printer/TextInstructionVisitor.swift
kateinoigakukun: Hmm, I do want to keep the most of the instruction info in the JSON file as the single source of truth. I think what this printer should be responsible is just for formatting operand and control flow stuff. We can just implement each instruction visit methods

Sources/WAT/Printer/ModuleCollector.swift
MaxDesiatov: The main problem is that name section is the last section, but we need that to arrive before we can emit any named declarations (which is all if not most of them?). Do you think it still makes sense to consider streaming here?

Sources/WAT/Printer/TextInstructionVisitor.swift
MaxDesiatov: That makes sense, I addressed it in https://github.com/swiftwasm/WasmKit/pull/305/changes/15604f7634ac2e56dbb2077919e776efc2c9a1ee

Sources/WAT/Printer/ModuleCollector.swift
kateinoigakukun: Hmm, good point. How about skimming all sections first without parsing contents, then parse dependent sections that we need to emit the rest of sections, then streamingly parse/emit the rest of sections?
My main concern for collecting everything as Swift data structure is that we need to model every Wasm entity in Swift even though we rarely manipulate them.

Sources/CLICommands/Wasm2wat.swift
```diff
@@ -0,0 +1,55 @@
...
+    var output: String?
+
+    package init() {}
+
+    package func run() throws {
+        let filePath = FilePath(path)
+        let fileHandle = try FileDescriptor.open(filePath, .readOnly)
+        try withThrowing {
+            let stream = try FileHandleStream(fileHandle: fileHandle)
+
```
MaxDesiatov:
```suggestion
```

Sources/WasmParser/Parser+Streaming.swift
```diff
@@ -0,0 +1,127 @@
+import WasmTypes
+
+extension Parser {
+    /// Parses one function-body encoding from the code section: a
+    /// size-in-bytes prefix, the locals-declaration vector, and the raw
+    /// expression bytes (including the trailing `end`). The returned
+    /// `Code.expression` is an `ArraySlice` of the parser's stream;
+    /// `Code.offset` is the absolute parent-buffer index where the
+    /// expression begins.
```
MaxDesiatov:
```suggestion
```

Sources/WasmParser/Parser+Streaming.swift
```diff
@@ -0,0 +1,128 @@
...
+        )
+        return Code(
+            locals: locals, expression: expressionBytes,
+            offset: expressionStart, features: features
+        )
+    }
+
+    /// Parses one data segment from the data section. Three encoding
+    /// kinds (0/1/2) are supported: active-into-memory-0, passive,
+    /// active-into-explicit-memory-index.
```
MaxDesiatov:
```suggestion
```

Sources/WasmParser/Parser+Streaming.swift
```diff
@@ -0,0 +1,128 @@
...
+            let index: UInt32 = try parseUnsigned()
+            let offset = try parseConstExpression()
+            let initializer = try parseVectorBytes()
+            return .active(.init(index: index, offset: offset, initializer: initializer))
+        }
+    }
+
+    /// Parses one element segment from the element section. The flag byte
+    /// encoding determines mode (active/passive/declarative), reference
+    /// type, and initializer shape.
```
MaxDesiatov:
```suggestion
```

Sources/WAT/Printer/ModuleCollector.swift
```diff
@@ -0,0 +1,183 @@
...
+    package var functionTypeIndices: [TypeIndex] = []
+    package var tables: [Table] = []
+    package var memories: [Memory] = []
+    package var globals: [Global] = []
+    package var exports: [Export] = []
+    package var start: FunctionIndex? = nil
+    /// Tags from the EH proposal's tag section.
+    package var tags: [Tag] = []
+
+    /// Raw byte slices for the big sections; sub-parsed during emit.
```
MaxDesiatov:
```suggestion
```

Sources/WAT/Printer/ModuleCollector.swift
```diff
@@ -0,0 +1,183 @@
...
+/// Collected information from all sections of a WebAssembly binary module.
+///
+/// "Big" sections (code, data, element) are captured as raw byte slices and
+/// streamed during emit. "Small" sections (type, import, function, table,
+/// memory, global, export, start, tag, dataCount) are eagerly materialised
+/// because their entries are bounded.
+package struct ModuleInfo {
+    package var types: [FunctionType] = []
+    package var imports: [Import] = []
+    /// Type indices from the function section (one per local function).
```
MaxDesiatov:
```suggestion
```

Sources/WAT/Printer/ModuleCollector.swift
```diff
@@ -0,0 +1,183 @@
...
+    package var codeSectionBytes: ArraySlice<UInt8>? = nil
+    package var dataSectionBytes: ArraySlice<UInt8>? = nil
+    package var elementSectionBytes: ArraySlice<UInt8>? = nil
+
+    /// DataCount value (when the data-count section is present); validated
+    /// against the data section's actual segment count during emit.
+    package var dataCount: UInt32? = nil
+
+    /// Threaded so the per-section sub-parsers in WatPrinter use the same
+    /// feature set as the original parse.
```
MaxDesiatov:
```suggestion
```

PR #322, opened 2026-03-03
Add annotations proposal implementation

MaxDesiatov:
`wasm-tools print` includes custom annotations on certain Wasm components, which WasmKit is currently unable to read.

Annotations proposal enables usage of custom sections and names in WAT that weren't previously expressible as WAT ids or strings. See [proposal overview](https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md) for more details.


Vendor/dependencies.json
```diff
@@ -1,4 +1,15 @@
...
+    "repository": "https://github.com/WebAssembly/spec.git",
+    "revision": "1c9a04e7b9d6998b9a30c6edf40b42d4c2a13c7b",
+    "categories": ["spec"]
+  },
+  "annotations": {
+    "repository": "https://github.com/WebAssembly/annotations.git",
+    "revision": "20a8e4f125d5d856aac392105c35055af9ed49f7",
+    "sparse-checkout": "test/custom",
+    "categories": ["default"]
+  },
```
kateinoigakukun: Do we need those repositories as dependencies?

Tests/WATTests/Spectest.swift
```diff
@@ -16,7 +16,7 @@ enum Spectest {
         testsuitePath.appendingPathComponent(file)
     }
 
-    static func wastFiles(include: [String] = [], exclude: [String] = ["annotations.wast"]) -> [URL] {
+    static func wastFiles(include: [String] = [], exclude: [String] = ["id.wast"]) -> [URL] {
```
kateinoigakukun, line 21: Why do we need to add `id.wast` as a new excluded test case?

Sources/WAT/BinaryEncoding/Encoder.swift
```diff
@@ -625,30 +641,48 @@ func encode(module: inout Wat, options: EncodeOptions) throws(WatParserError) ->
...
+    emitCustomSections(placement: .before(.import), encoder: &encoder)
     if !module.imports.isEmpty {
         encoder.section(id: 0x02) { encoder in
             encoder.encodeVector(module.imports)
         }
     }
+    emitCustomSections(placement: .after(.import), encoder: &encoder)
 
     // Section 3: Function section
+    emitCustomSections(placement: .before(.func), encoder: &encoder)
```
kateinoigakukun: It's very error-prune to enforce inserting emitCustomSections before/after each section manually. Can we make it as a part of `Encoder.section`?

Sources/WAT/Lexer.swift
```diff
@@ -37,6 +40,14 @@ enum IntegerToken: Equatable {
 struct Token {
     let range: Range<Lexer.Index>
     let kind: TokenKind
+    /// For quoted identifiers ($"..."), the decoded string bytes (not including the `$` prefix).
+    let quotedIdBytes: [UInt8]?
```
kateinoigakukun: Why do we need to hold it here? Can't we simply interpret it in Parser?

Sources/WAT/Lexer.swift
```diff
@@ -328,6 +367,112 @@ struct Lexer {
         }
     }
 
+    /// Recognized annotation IDs that the parser needs to handle.
+    private static let recognizedAnnotations: Set<String> = ["name", "custom"]
```
kateinoigakukun: I don't think Lexer needs to know about them. We can treat all custom section annotation items as opaque in Lexer, and parse known custom sections in Parser when needed. We can remove `skipAnnotationBody`.

kateinoigakukun (commented):
I feel like this change is too vibey 😞

Vendor/dependencies.json
MaxDesiatov: Only these repositories have relevant WAST tests for the proposal. We could use a sparse checkout to pull in just the few WAST files if you prefer that?

kateinoigakukun: Ok, the root problem here is that those wast files are not included in testsuite repository. I've submitted a fix now https://github.com/WebAssembly/testsuite/pull/155

Also do we still need spec repository?

Tests/WATTests/Spectest.swift
MaxDesiatov: `id.wast` needed to be excluded because this function runs against WABT/WasmTools baseline, and neither of those support this WAST file, it's tested separately in `func annotationProposalSpectest()`, and that also has an explanation comment

Vendor/dependencies.json
kateinoigakukun: The testsuite PR has been merged

MaxDesiatov: Unfortunately, the PR has been merged post Wasm-3.0 additions, which merged all proposals into the main test suite directory, which means we can't easily exclude GC tests directory wholesale if we bump to the commit where this PR is merged.

I'm vendoring the new tests for now.

Sources/WAT/Lexer.swift
MaxDesiatov: Thanks, that's cleaned up.

Sources/WAT/Lexer.swift
MaxDesiatov: That's addressed on the `Parser` level now.

Sources/WAT/BinaryEncoding/Encoder.swift
MaxDesiatov: Makes sense, done.

PR #327, opened 2026-03-04
Fix fd handling in WASIp1 implementation

### LINKED ISSUE #334: Unable to run CPython in WasmKit
Steps to reproduce:
1. Build CPython 3.14 for `wasm32-wasip1` (any recent version)
2. Run `python.wasm` with WasmKit (0.2.1 or main), mounting the stdlib as a preopen:
```
wasmkit-cli run --dir /path/to/lib/python3.14 \
  --env PYTHONHOME=/path/to \
  --env PYTHONDONTWRITEBYTECODE=1 \
  -- python.wasm -c "print('hello')"
```

Expected: Python prints `hello`

Actual: Crash with:
```
Fatal Python error: Failed to import encodings module
Python runtime state: core initialized
Error: Bad file descriptor
```

MaxDesiatov:
`PathResolution.resolve()` returned the input fd unchanged for "." paths, causing fd table entries to alias the same host fd. Closing one invalidated the other, which errored with EBADF.

Multiple related fixes are included to clean up the general situation with FD handling in WASIp1 implementation:

- Fix fd aliasing in `PathResolution.resolve()`: When the path resolves without opening a new fd (e.g. "."), dup via `openat(fd, ".")` so callers always get an independently closeable fd.
- Fix `SplitPath` premature deinit: Use `borrowing func withFields` to keep `SplitPath: ~Copyable` alive during operations that use its parent fd.
- Add `WASIBridgeToHost.runAndClose`: Scoped API that ensures `close()` is called, replacing the previous `deinit`-based cleanup.
- Add `isBorrowed` to `WASIEntry`: Stdio entries are marked as borrowed so `closeAll()` never closes process stdin/stdout/stderr (previously broken when `fd_renumber` moved stdio to different WASI fd positions).
- Fix `directoryEntry(fd:)` error codes: Return `EBADF` for missing fds, `ENOTDIR` for non-directory fds (was `ENOTDIR` for both).

I ran tests locally 10 times in a row and on CI 5 times in a row on this PR, no failures, I think this is sufficient to say that flakiness is resolved.

Resolves https://github.com/swiftwasm/WasmKit/issues/334.

### DIFF
--- a/Sources/WASI/WASIBridgeToHost.swift
+++ b/Sources/WASI/WASIBridgeToHost.swift
@@ -14,8 +14,9 @@ import SystemPackage
 ///     environment: ["PATH": "/usr/bin"],
 ///     preopens: [WASIBridgeToHost.Preopen(guestPath: "/sandbox", hostPath: "/real/path")]
 /// )
-/// // ... use bridge ...
-/// try bridge.close()
+/// try bridge.runAndClose { wasi in
+///     // ... use wasi ...
+/// }
 /// ```
 public final class WASIBridgeToHost {
     internal let underlying: WASIImplementation
@@ -111,6 +112,22 @@ public final class WASIBridgeToHost {
     public func close() throws {
         try underlying.close()
         isClosed = true
+    }
+
+    /// Passes the bridge to `body`, then closes all owned file descriptors
+    /// regardless of whether `body` returns normally or throws.
+    ///
+    /// - Parameter body: A closure that receives the bridge and returns a value.
+    /// - Returns: The value returned by `body`.
+    public func runAndClose<R>(_ body: (WASIBridgeToHost) throws -> R) throws -> R {
+        do {
+            let result = try body(self)
+            try close()
+            return result
+        } catch {
+            try close()
+            throw error
+        }
     }
 
     deinit {
--- a/Sources/WASI/WASIBridgeToHost.swift
+++ b/Sources/WASI/WASIBridgeToHost.swift
@@ -169,7 +169,6 @@ public final class WASIBridgeToHost {
     ///   - monotonicClock: Clock for monotonic time queries. Defaults to `SystemMonotonicClock()`.
     ///   - randomGenerator: Random number generator. Defaults to `SystemRandomNumberGenerator()`.
     /// - Throws: An error if the file system or preopens cannot be initialized.
-    @available(*, deprecated, message: "Use withBridge(args:environment:fileSystem:body:) instead to ensure file descriptors are properly closed.")
     public convenience init(
         args: [String] = [],
         environment: [String: String] = [:],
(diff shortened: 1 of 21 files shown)


Sources/WASI/WASIBridgeToHost.swift
```diff
@@ -162,6 +169,7 @@ public final class WASIBridgeToHost {
     ///   - monotonicClock: Clock for monotonic time queries. Defaults to `SystemMonotonicClock()`.
     ///   - randomGenerator: Random number generator. Defaults to `SystemRandomNumberGenerator()`.
     /// - Throws: An error if the file system or preopens cannot be initialized.
+    @available(*, deprecated, message: "Use withBridge(args:environment:fileSystem:body:) instead to ensure file descriptors are properly closed.")
```
kateinoigakukun: I think enforcing `with` scoped API is too restrictive here because any Wasm instance cannot outlive the scope but we often need it.

MaxDesiatov:
```suggestion
```

Sources/WASI/WASIBridgeToHost.swift
```diff
@@ -196,24 +223,56 @@ public final class WASIBridgeToHost {
     ///   - monotonicClock: Clock for monotonic time queries. Defaults to `SystemMonotonicClock()`.
     ///   - randomGenerator: Random number generator. Defaults to `SystemRandomNumberGenerator()`.
     /// - Throws: An error if the file system or initialization fails.
-    public init(
+    @available(*, deprecated, message: "Use withBridge(args:environment:fileSystem:body:) instead to ensure file descriptors are properly closed.")
```
MaxDesiatov:
```suggestion
```

Sources/WASI/WASIBridgeToHost.swift
MaxDesiatov: I actually agree now, previously I did try to convert more code to the `with`-function and couldn't do it everywhere, but I forgot to remove deprecations in this PR before requesting a review, sorry about that. The deprecation attributes are removed now.

Sources/WASI/WASIBridgeToHost.swift
```diff
@@ -196,24 +222,55 @@ public final class WASIBridgeToHost {
...
+            args: args, environment: environment, fileSystemOptions: fileSystem,
+            wallClock: wallClock, monotonicClock: monotonicClock,
+            randomGenerator: randomGenerator
+        )
+        do {
+            let result = try body(bridge)
+            try bridge.close()
+            return result
+        } catch {
+            try? bridge.close()
```
kateinoigakukun: Hmm, ignoring the fd close errors doesn't sound very good. I'm not very confident about the withBridge surface. Can we remove it for now?

MaxDesiatov: I think there's still value in something like `runAndClose` helper that I added as instance method in [c7ec912](https://github.com/swiftwasm/WasmKit/pull/327/commits/c7ec91288d3938211083776be81be9db5c02f791), because `defer { try wasi.close() }` won't work (trying `defer` is not possible in Swift), so this incentivises people to write `defer { try? wasi.close() }`.

kateinoigakukun (approved):

PR #328, opened 2026-03-10
Add `Caller.captureBacktrace()`

pcbeard:
This allows a Wasm host function to capture a back trace, which can be useful for debugging.

### DIFF
--- a/Sources/WasmKit/Execution/Store.swift
+++ b/Sources/WasmKit/Execution/Store.swift
@@ -36,6 +36,7 @@ extension Store: Equatable {
 }
 
 /// A caller context passed to host functions
+/// Not copyable to avoid storing stale stack pointer.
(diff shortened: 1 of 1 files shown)


Sources/WasmKit/Execution/Store.swift
```diff
@@ -36,8 +36,10 @@ extension Store: Equatable {
 }
 
 /// A caller context passed to host functions
-public struct Caller {
+public struct Caller: ~Copyable {
```
kateinoigakukun, line 40: Let's add a rationale doc comment for the `~Copyable`

pcbeard: Done.

kateinoigakukun (approved):
Thanks!

PR #329, opened 2026-03-11
Add DWARF line table parser for source-level backtraces

pcbeard:
Parse DWARF .debug_line, .debug_info, .debug_abbrev, and .debug_ranges sections from WASM custom sections to resolve code addresses to source file/line/column locations. Used by Caller.captureBacktrace() to provide source-level information in stack traces.

Key components:
- DWARFLineTable: parses DWARF v4/v5 line number programs with CU-aware resolution using .debug_info address ranges
- Code section offset conversion between file-level and DWARF addresses
- binaryOffsetCorrection in Translator to fix buffer-relative offsets
- Backtrace.Symbol.sourceLocation for resolved source locations
- Precise line resolution via instruction mapping with function-start fallback
- Translator fix from expose-debugging-api for folded instruction mapping
- codeSectionStart exposed via ParsingPayload.codeSection
- 13 unit tests for the DWARF parser

### DIFF
--- a/Sources/WasmKit/Execution/DWARFLineTable.swift
+++ b/Sources/WasmKit/Execution/DWARFLineTable.swift
@@ -716,14 +716,15 @@ private struct Cursor {
 
     mutating func readU64() throws -> UInt64 {
         guard offset + 8 <= data.endIndex else { throw DWARFError.unexpectedEnd }
-        let value = UInt64(data[offset])
-            | (UInt64(data[offset + 1]) << 8)
-            | (UInt64(data[offset + 2]) << 16)
-            | (UInt64(data[offset + 3]) << 24)
-            | (UInt64(data[offset + 4]) << 32)
-            | (UInt64(data[offset + 5]) << 40)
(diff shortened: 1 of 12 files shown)


Sources/WasmKit/Execution/DWARFLineTable.swift
MaxDesiatov, line 1: There's ongoing work on offline symbolication for Swift on other platforms. If we could agree on a format that WasmKit outputs, maybe the symbolicator could accept that and we wouldn't need DWARF parsing in WasmKit at all

pcbeard: Could be optional of course, but handy during debugging. Still need something to decode the translated PC values to correct offsets that can be used offline. I don't believe the current PC values would work.

MaxDesiatov: I'd be happy to adjust values to whatever symbolicator is expecting. Probably also need to align WasmKit with output from other Wasm runtimes so that the symbolicator can expect consistent input.

PR #341, opened 2026-04-28
Fix subscript visibility

slavapestov:
The nominal type has package visibility, so the dynamicMember subscript must also have package visibility.

This is now diagnosed with https://github.com/swiftlang/swift/pull/88480.

### DIFF
--- a/Sources/WasmKit/Execution/Instances.swift
+++ b/Sources/WasmKit/Execution/Instances.swift
@@ -51,7 +51,7 @@ package struct EntityHandle<T>: Equatable, Hashable {
         self.pointer = pointer
     }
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Would you mind providing a description for a bit more context?

Also, would need a PR against `main` as well?

slavapestov:
I updated the PR description, does it make sense?

It looks like the subscript was already changed to `package` visibility on `main`, so no need to fix anything there.

I might need to downgrade this to a warning anyway, in which case I'll close this PR.

kateinoigakukun (approved):
Thanks, it makes sense to me.

PR #343, opened 2026-05-10
docs: add usage guides for WITOverlayGenerator and WITExtractor

shivanshu877:
Addresses the two open items in https://github.com/swiftwasm/WasmKit/issues/71 (\"Enhance documentation\"):

- Example and usage of `WITOverlayGenerator`
- Example and usage of `WITExtractor`

Both documents are derived from reading the source of the respective modules (`WITOverlayGen.swift`, `WITExtractor.swift`) and their SPM plugins (`WITOverlayPlugin`, `WITExtractorPlugin`).

The intent is the same as the existing module docs in WasmKit's docc bundle — make the WIT tooling discoverable for users coming from outside the project.

## Test plan

- [x] Docs build cleanly with the existing docc setup.
- [ ] A maintainer review on the prose / accuracy is welcome — I worked from the source, but anyone closer to the API will see corner cases I missed.

### DIFF
--- a/Documentation/ComponentModel/WITExtractor.md
+++ b/Documentation/ComponentModel/WITExtractor.md
@@ -1,0 +1,182 @@
+# WITExtractor — Usage Guide
+
+`WITExtractor` derives a [WIT](https://component-model.bytecodealliance.org/design/wit.html) (WebAssembly Interface Types) interface description from an **existing Swift module**. It is the reverse of `WITOverlayGenerator`: you start with Swift source and get a `.wit` file plus a Swift overlay that exports your module as a Wasm component.
+
+## When to use it
+
+Use `WITExtractor` when you have a Swift library you want to expose as a WebAssembly component but you do not want to maintain a hand-written WIT file. The extractor introspects the compiled module and generates both:
+
+1. A `.wit` file — the component interface other languages can consume.
+2. A Swift overlay `.swift` file — Canonical-ABI entry points that export your Swift functions to the component model host.
+
+## How it works
+
+`WITExtractor` shells out to `swift-api-digester` (bundled with the Swift toolchain) to obtain a JSON description of the module's public API. It then:
+
+1. **Collects types** — maps Swift types to WIT types (see table below).
+2. **Builds a source summary** — collects public functions, structs, and enums.
+3. **Translates to WIT** — emits a `package <namespace>:<package-name>` block containing a single interface named after the module.
+4. **Generates a Swift overlay** — emits `@_expose(wasm, ...)` entry points that lift/lower Canonical-ABI values.
+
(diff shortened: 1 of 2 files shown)


Documentation/ComponentModel/WITExtractor.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+---
+
+## Usage via `wit-tool` CLI
+
+For scripted use, call the `extract-wit` subcommand directly:
+
+```sh
+swift run wit-tool extract-wit \
```
MaxDesiatov: It doesn't look like these instructions were verified end-to-end, or at least not verified with the state of the repository as I see it? `swift run wit-tool` won't work. This will fail with `error: no executable product named wit-tool`


```suggestion
swift run WITTool extract-wit \
```

Documentation/ComponentModel/WITExtractor.md
```diff
@@ -0,0 +1,182 @@
...
+
+---
+
+## End-to-end example
+
+Given a Swift module `MathLib` with the following public API:
+
+```swift
+// Sources/MathLib/MathLib.swift
+public struct Vector2 {
```
MaxDesiatov, line 135: IIRC `WITExtractor` only extracts decls with `@_spi(WIT)` on them.

Documentation/ComponentModel/WITExtractor.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Usage via `wit-tool` CLI
+
+For scripted use, call the `extract-wit` subcommand directly:
+
+```sh
+swift run wit-tool extract-wit \
+    --swift-api-digester $(xcrun --find swift-api-digester) \
+    --module-name MyLibrary \
+    --package-name my-library \
```
MaxDesiatov, line 109: the plugin form only accepts `--target`, `--sdk`, `--output-mapping`. It auto-derives the WIT package name from `context.package.displayName`. This snippet conflates the plugin and the CLI

MaxDesiatov (changes requested):

shivanshu877:
Thanks @MaxDesiatov — all three addressed in `ef224eb`:

1. **`swift run wit-tool` → `swift run WITTool`** (verified `WITTool` is the actual executable name in `Package.swift`).
2. **End-to-end example no longer conflates plugin and CLI.** Reworked the "Running…" snippet to use the plugin form (`swift package plugin extract-wit --target MathLib`) with a parenthetical pointing at the CLI form for callers that need to override the package name.
3. **Documented the `@_spi(WIT)` opt-in requirement.** Added a callout in the "When to use it" section explaining only `@_spi(WIT)`-annotated declarations are extracted, and applied `@_spi(WIT)` to the example's `Vector2` and `addVectors` so the snippet actually produces the WIT shown below it (verified against `Sources/WITExtractor/SourceSummary.swift` line 221: `guard let spiGroup = decl.body.spi_group_names, spiGroup.contains(\"WIT\") else { return false }`).

Ready for re-review.

PR #359, opened 2026-06-23
Add shared linear memory as prerequisite for `wasip1-threads`

MaxDesiatov:
Added corresponding shared memory load/instructions that call `reloadSharedMemorySize`, updated atomic instructions to do the same.

### DIFF
--- a/Sources/WasmKit/Execution/Execution.swift
+++ b/Sources/WasmKit/Execution/Execution.swift
@@ -513,7 +513,13 @@ extension Execution {
             var pc = pc
             var md = md
             var ms = ms
-            let shouldUseMprotectTrapGuards = store.value.engine.configuration.memoryBoundsChecking == .mprotect
+            // Shared memory always uses software bounds checking. A guard-page fault
+            // recovered by siglongjmp would unwind the interpreter past a lock held by a
+            // concurrent grow or atomic operation, so the multi-threaded path must stay
+            // free of non-local jumps. mprotect guards are reserved for non-shared memory.
+            let isNonSharedMemory = sp.currentInstance?.memories.first?.withValue { $0.sharedStorage == nil } ?? true
+            let shouldUseMprotectTrapGuards =
+                store.value.engine.configuration.memoryBoundsChecking == .mprotect && isNonSharedMemory
             let storeValue = store.value
             while true {
                 let handler = pc.read(wasmkit_tc_exec.self)
--- a/Sources/WasmKit/Execution/Instances.swift
+++ b/Sources/WasmKit/Execution/Instances.swift
@@ -66,6 +66,30 @@ package struct EntityHandle<T: ~Copyable>: Equatable, Hashable, Copyable {
     }
 }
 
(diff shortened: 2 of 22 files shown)


Sources/WasmKit/Execution/Execution.swift
```diff
@@ -513,7 +513,13 @@ extension Execution {
...
             var md = md
             var ms = ms
-            let shouldUseMprotectTrapGuards = store.value.engine.configuration.memoryBoundsChecking == .mprotect
+            // Shared memory always uses software bounds checking. A guard-page fault
+            // recovered by siglongjmp would unwind the interpreter past a lock held by a
+            // concurrent grow or atomic operation, so the multi-threaded path must stay
+            // free of non-local jumps. mprotect guards are reserved for non-shared memory.
+            let isNonSharedMemory = sp.currentInstance?.memories.first?.withValue { $0.sharedStorage == nil } ?? true
+            let shouldUseMprotectTrapGuards =
+                store.value.engine.configuration.memoryBoundsChecking == .mprotect && isNonSharedMemory
```
kateinoigakukun, line 522: Hmm, I actually didn't get this. Is it even theoretically possible that "thread X causes Wasm OOB access while owning memory.grow or atomic.wait lock? Given that `siglongjmp` is a thread local stack unwinding, I guess it won't be possible to happen?

Sources/WasmKit/Execution/StoreAllocator.swift
```diff
@@ -155,50 +156,59 @@ extension ImmutableArray: Sequence {
 
 /// A type that can be interned into a unique identifier.
 /// Used for efficient equality comparison.
-protocol Internable {
+package protocol Internable {
```
kateinoigakukun: Q: Why do we need to expose some structures in this file to `package` level?

Sources/WasmKit/Execution/Instances.swift
```diff
@@ -589,7 +613,10 @@ struct MemoryEntity: ~Copyable {
     private var storage: Storage
     let maxPageCount: UInt64
     let limit: Limits
-    let sharedMutex: Mutex<Void>?
+
+    /// If this memory is shared, the mmap-backed storage is held here.
+    /// Multiple `MemoryEntity` instances (across threads) may reference the same `SharedMemoryStorage`.
+    let sharedStorage: SharedMemoryEntity?
```
kateinoigakukun, line 619: I think the memory storage should be consistently represented by `let storage` member to avoid having many `if let sharedStorage` below and we can remove shared-variant of memory instructions. Also even further, I'm wondering if we really need a new `SharedMemoryEntity` type here, can't we add shared mode to `MprotectLinearMemory`?

kateinoigakukun (changes requested):
Let's split out AtomicParkingLot's `pthread_cond_t` change to make review easier 🙇

MaxDesiatov:
Closing as superseded by https://github.com/swiftwasm/WasmKit/pull/380

PR #363, opened 2026-06-27
WASI: Fix fd leak in `fd_readdir`

### LINKED ISSUE #362: Suspected File Descriptor Leak
It appears that the latest WasmKit (a0010b6eb445d87b26942466a574e1e8202e2cdb) still has a file descriptor leak issue. I inferred that from the results of `ParserTests.testSwiftTestsuite` in swiftlang/swift-syntax.

- https://github.com/swiftlang/swift-syntax/issues/3369 (Please see the "Minor problems" section)

That test function finds files with `FileManager.enumerator(at:includingPropertiesForKeys:)` and reads 6920 files in parallel. (However, as you know, it is not actually parallel in Wasm.)

I added a test case that checks file descriptor leak to WasmKit before.

https://github.com/swiftwasm/WasmKit/blob/a0010b6eb445d87b26942466a574e1e8202e2cdb/Tests/WASITests/WASITests.swift#L1564

However, since it did not actually compare the number of file descriptors, it was an incomplete test.

kateinoigakukun:
We didn't call `closedir` corresponding to the `fdopendir` call, which caused file descriptor leaks when `fd_readdir` was called.

Close https://github.com/swiftwasm/WasmKit/issues/362

<details>
<summary>Investigation notes</summary>

1. Reproduced the failure locally by lowering the per-process fd limit.
2. Let agent instrument WasmKit to trace:
   - WASI syscall entry/exit for `path_open` and `fd_close`
   - host-side open/close points
3. Ran the failing swift-syntax test with tracing enabled and compared open/close counts.
4. Observed that:
   - guest `path_open` and `fd_close` were balanced
   - repeated `fd_readdir` calls caused host-side duplicated directory handles
   - those handles were not being closed
5. Followed the trace to the `fdopendir` path and identified the missing `closedir` as the leak point.
</details>

### DIFF
--- a/Sources/SystemExtras/FileOperations.swift
+++ b/Sources/SystemExtras/FileOperations.swift
@@ -571,6 +571,10 @@ extension FileDescriptor {
       self.rawValue = rawValue
     }
 
+    public func close() {
+      _ = system_closedir(rawValue)
+    }
+
     public func next() -> Result<DirectoryEntry, Errno>? {
       // https://man7.org/linux/man-pages/man3/readdir.3.html#RETURN_VALUE
       // > If the end of the directory stream is reached, NULL is returned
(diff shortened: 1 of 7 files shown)


MaxDesiatov (approved):
Thanks! You beat me to it, I was just creating a PR for this 😅

PR #380, opened 2026-07-20
Add shared memory support for mprotect-able platforms

kateinoigakukun:
Adds support for shared linear memory (the WebAssembly threads proposal), building on #359 but dropping software-bounds-checking support for shared memory to keep the approach small.

Atomic instructions already work on `main`. This makes a `shared` memory instantiable and shareable across threads, including cross-thread `memory.atomic.wait`/`notify`.
ndows).

## Limitations

Shared memory requires mprotect bounds checking (64-bit macOS/Linux). Under software bounds checking, token threading, ASan, Windows, WASI, or memory64, creating a shared memory traps with `sharedMemoryRequiresMprotect`.


Sources/WasmKit/Execution/Instances.swift
```diff
@@ -506,104 +506,209 @@ struct MemoryEntity: ~Copyable {
...
-    #if os(macOS) || os(Linux)
-        private enum Storage {
+    /// Backing storage for a linear memory; shared memories are the `.shared` case.
+    private enum Storage {
+        #if os(macOS) || os(Linux)
             case mprotect(MprotectLinearMemory)
-            case malloc(MallocStorage)
+        #endif
+        case malloc(MallocStorage)
+        case shared(SharedMemoryStorage)
```
MaxDesiatov: Would it structurally make more sense to move this `case` within the `#if` condition?

```suggestion        
            case shared(SharedMemoryStorage)
        #endif
        case malloc(MallocStorage)
```

Sources/WasmKit/Execution/Instances.swift
```diff
@@ -506,104 +506,209 @@ struct MemoryEntity: ~Copyable {
...
-        self.storage = Storage(byteSize: byteSize, isMemory64: memoryType.isMemory64, engineConfiguration: engineConfiguration)
+
         let defaultMaxPageCount = Self.maxPageCount(isMemory64: memoryType.isMemory64)
-        maxPageCount = memoryType.max ?? defaultMaxPageCount
+        let maxPages = memoryType.max ?? defaultMaxPageCount
+        // `max` is only a growth ceiling, not an amount we allocate. If it overflows the
+        // representable range (e.g. a large memory64 maximum), clamp it to the largest
+        // page-aligned value instead of failing: the real cap is the reservation / the host
+        // address space, enforced at grow time. This mirrors wasmtime, which silently clamps
+        // the effective maximum (only the *minimum* overflowing is an instantiation error).
```
MaxDesiatov: This feels redundant as it rehashes what Wasm spec already states?

MaxDesiatov (approved):
Overall LGTM modulo some redundant comments and the fact that maybe shared memory case code can be hidden under `#if` condition to reduce code size on platforms that don't support it?

PR #381, opened 2026-07-21
SwiftSyntax-based `WITExtractor` re-implementation

MaxDesiatov:
`swift-api-digester` approach is no longer available with SwiftBuild due to significant changes in what's available to plugins and how modules are laid out on the file system.

In this PR I propose replacing it with a SwiftSyntax approach similar to what JavaScriptKit does already.


Tests/WITExtractorPluginTests/TestSupport.swift
```diff
@@ -1,27 +1,12 @@
...
-        }()
-    }
-
     static func withTemporaryDirectory<Result>(
         _ body: (String) throws -> Result
     ) throws -> Result {
         let tempdir = URL(fileURLWithPath: NSTemporaryDirectory())
         let templatePath = tempdir.appendingPathComponent("WasmKit.XXXXXX")
-        var template = [UInt8](templatePath.path.utf8).map({ UInt8($0) }) + [UInt8(0)]
+        var template = [UInt8](templatePath.path.utf8).map({ Int8($0) }) + [Int8(0)]
```
MaxDesiatov: This and a few other lines are fixes for "warnings as errors" now enabled on these modules.

Sources/WITMarker/WITMarker.swift
```diff
@@ -0,0 +1,5 @@
+/// Marks a declaration for extraction into a WIT interface by the WITExtractor tool.
+///
+/// Expands to nothing; WITExtractor detects `@WIT` by parsing source, not expansion.
+@attached(peer)
+public macro WIT() = #externalMacro(module: "WITMacroPlugin", type: "WITMacro")
```
kateinoigakukun: https://github.com/swiftwasm/JavaScriptKit/blob/0e35c39017edfa6a18d2570bdfc06ebfdcca73be/Sources/JavaScriptKit/Macros.swift#L116

This is very hacky but can we use the same way we use in BridgeJS to save time for building a macro plugin?

Package.swift
```diff
@@ -1,5 +1,6 @@
 // swift-tools-version:6.3
 
+import CompilerPluginSupport
```
kateinoigakukun:
```suggestion
```

kateinoigakukun (approved):

PR #382, opened 2026-07-21
Fix debugger to attribute breakpoint stops correctly

MaxDesiatov:
The handler reported every stop as `reason:trace` at the resolved instruction, so LLDB mapped hits to `eStopReasonTrace` and never attributed them to a breakpoint. It now tracks each host-requested address, replies with `reason:breakpoint` at that address on a real hit while keeping step and entrypoint landings as `trace`, and removes breakpoints at their resolved address. Added a `WasmKitGDBHandlerTests` target covering the stop reason, reported address, and removal.

### DIFF
--- a/Sources/WasmKitGDBHandler/WasmKitGDBHandler.swift
+++ b/Sources/WasmKitGDBHandler/WasmKitGDBHandler.swift
@@ -60,6 +60,11 @@
         private var memoryView: DebuggerMemoryView
         private let wasi: WASIBridgeToHost
 
+        // Requested and resolved addresses diverge when the requested address lands on an elided
+        // instruction (e.g. local.get) that resolves forward; the host expects stop replies at the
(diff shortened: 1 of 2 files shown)


Sources/WasmKitGDBHandler/WasmKitGDBHandler.swift
```diff
@@ -59,6 +59,11 @@
         private var memoryView: DebuggerMemoryView
         private let wasi: WASIBridgeToHost
 
+        // Requested and resolved addresses diverge when the requested address lands on an elided
+        // instruction (e.g. local.get) that resolves forward; the host expects stop replies at the
+        // address it set the breakpoint on, not the resolved one.
+        private var resolvedToRequestedBreakpoint = [Int: Int]()
```
kateinoigakukun, line 66: Hmm, this smells a bit. Don't we already track the same state in Debugger?

PR #383, opened 2026-07-21
Basic multi-memory proposal support

MaxDesiatov:
Add the WebAssembly multi-memory proposal as an opt-in WasmFeatureSet feature. `memory.init`/`copy`/`fill` now carry an explicit memory index through the binary decoder, WAT text parser and encoder, instruction visitor, translator, and runtime; the validator permits multiple memories when `.multiMemory` is set. `memory.copy` supports two distinct memories with per-memory index-type widening. The VM instruction spec is regenerated via VMGen for the new memory-index operands.

Enabling multi-memory for the vendored spec-test suite (which needs a testsuite pin bump) is deferred; default parsing and validation behavior is unchanged.

### DIFF
--- a/Sources/WAT/BinaryEncoding/Encoder.swift
+++ b/Sources/WAT/BinaryEncoding/Encoder.swift
@@ -503,12 +503,6 @@ struct ExpressionEncoder: BinaryInstructionEncoder {
     }
 
     // MARK: Special instructions
-    mutating func visitMemoryInit(dataIndex: UInt32) {
-        encodeInstruction([0xFC, 0x08])
-        encodeImmediates(dataIndex: dataIndex)
-        encodeByte(0x00)  // reserved value
-    }
-
     mutating func visitTypedSelect(type: ValueType) {
         encodeInstruction([0x1C])
         encodeByte(0x01)  // number of result types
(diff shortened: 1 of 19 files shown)


Sources/WasmParser/WasmParser.swift
```diff
@@ -665,28 +672,24 @@ extension Parser: BinaryInstructionDecoder {
...
+        return (destination, source)
     }
 
     @inlinable mutating func visitMemoryFill() throws(WasmParserError) -> UInt32 {
-        let zero = try stream.consumeAny()
-        guard zero == 0x00 else {
-            throw makeError(.zeroExpected(actual: zero))
-        }
-        return 0
+        try parseMemoryIndex()
```
kateinoigakukun, line 692: Should we gate behind the feature flag like what you do for visitMemoryInit?

Sources/WAT/BinaryEncoding/Encoder.swift
```diff
@@ -532,6 +526,12 @@ struct ExpressionEncoder: BinaryInstructionEncoder {
         hasDataSegmentInstruction = true
         encodeUnsigned(dataIndex)
     }
+    mutating func encodeImmediates(dataIndex: UInt32, memory: UInt32) {
+        // For memory 0, encoding the memory index yields the single 0x00 the pre-multi-memory reserved byte produced.
+        hasDataSegmentInstruction = true
+        encodeUnsigned(dataIndex)
```
kateinoigakukun, line 532:
```suggestion
        encodeImmediates(dataIndex: dataIndex)
```

Sources/WAT/Parser/WastParser.swift
```diff
@@ -327,13 +328,16 @@ public struct ModuleDirective {
     public let id: String?
     /// The location of the module in the source
     public let location: Location
+    /// The `(module definition ...)` form decodes and validates without instantiating.
+    public let isModuleDefinition: Bool
```
kateinoigakukun, line 332: Is this related to multi-memory proposal?

PR #357, opened 2026-06-22
Embedded Swift support

Joannis:

PR #391, opened 2026-07-27
Add printable core types and a typed Trap kind in WasmKitError

kateinoigakukun:
Part of the Embedded Swift support work (#357), split out for incremental review. Two commits:

**Add CustomStringConvertible to core Wasm types** — hand-written descriptions for `ValueType`, `ReferenceType`/`HeapType`, `FunctionType`, `Value`, `Reference`, `V128` (WasmTypes) and `Limits`, `TableType`, `GlobalType`, `Mutability` (WasmParser). String interpolation of these types previously fell back to reflection-based stringification, which Embedded Swift lacks; the explicit descriptions produce wasm-flavored text instead, e.g. `(i32, i32) -> (i32)`, `i32(5)`, `funcref`. Array formatting goes through static `descriptionList(_:)` methods on the element types rather than extensions on `Array` itself.

**WasmKitError: carry Trap as a typed error kind** — groundwork for typed throws in the instantiation path. `WasmKitError` gains a `trap` kind so `Trap`-throwing sub-operations can participate in a `throws(WasmKitError)` chain, `Trap`/`TrapReason` are declared `Sendable`, and the few reflection-dependent error messages (value arrays, instance handles, backtrace addresses) are rewritten into explicit formatting so they also compile under Embedded Swift. Error message text changes slightly where reflection dumps were previously used; no test relied on the old text.

### DIFF
--- a/Sources/WasmTypes/Description.swift
+++ b/Sources/WasmTypes/Description.swift
@@ -1,0 +1,117 @@
+// Human-readable descriptions for core Wasm types.
+//
+// These are hand-written (rather than relying on the reflection-based default
+// stringification) so that string interpolation of these types also works
+// under Embedded Swift, which has no reflection support.
+
+extension AbstractHeapType: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case .funcRef: return "func"
+        case .externRef: return "extern"
+        case .exnRef: return "exn"
+        }
+    }
+}
+
+extension HeapType: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case .abstract(let type): return type.description
+        case .concrete(let typeIndex): return "$\(typeIndex)"
+        }
+    }
+}
+
+extension ReferenceType: CustomStringConvertible {
+    public var description: String {
+        if isNullable {
+            switch heapType {
+            case .abstract(.funcRef): return "funcref"
+            case .abstract(.externRef): return "externref"
+            case .abstract(.exnRef): return "exnref"
+            case .concrete: break
+            }
+        }
+        return "(ref \(isNullable ? "null " : "")\(heapType))"
+    }
+}
+
+extension ValueType: CustomStringConvertible {
+    public var description: String {
+        switch self {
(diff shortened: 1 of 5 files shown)


Sources/WasmTypes/Description.swift
```diff
@@ -0,0 +1,111 @@
...
+    /// Returns the descriptions of `values` joined with ", ".
+    public static func descriptionList(_ values: [Value]) -> String {
+        values.map { $0.description }.joined(separator: ", ")
+    }
+}
+
+private func hexByte(_ byte: UInt8) -> String {
+    let digits = "0123456789abcdef"
+    let high = digits[digits.index(digits.startIndex, offsetBy: Int(byte >> 4))]
+    let low = digits[digits.index(digits.startIndex, offsetBy: Int(byte & 0xF))]
```
MaxDesiatov: Could we maybe have `digits` as an array of ASCII bytes that you can directly index with O(1)?

Sources/WasmTypes/Description.swift
```diff
@@ -0,0 +1,111 @@
...
+    /// Returns the descriptions of `types` joined with ", ".
+    public static func descriptionList(_ types: [ValueType]) -> String {
+        types.map { $0.description }.joined(separator: ", ")
+    }
+}
+
+extension Value {
+    /// Returns the descriptions of `values` joined with ", ".
+    public static func descriptionList(_ values: [Value]) -> String {
+        values.map { $0.description }.joined(separator: ", ")
```
MaxDesiatov, line 108: unfortunate that we don't have `description(into: inout UTF8Span)` or smth like that, printing values could potentially be a hot path after all (in debugging scenarios at least)

MaxDesiatov: hm, `UTF8Span` is immutable? Maybe `MutableSpan<UInt8>` or whatever, and we can assume `Value` has a max description length that we can statically preallocate a buffer for, v128 being the longest possible value

PR #392, opened 2026-07-27
Add `--argv0` and `--dir host::guest` mapping to `wasmkit run`

MaxDesiatov:
`wasmkit run` could only grant a host directory under its own path and always passed the module path as argv[0]. `--dir` now takes an optional `host::guest` mapping and `--argv0` overrides `argv[0]`, which is what CPython's `python.wasm` needs. A failed preopen now releases what it opened instead of tripping `WASIBridgeToHost`'s close precondition.

### DIFF
--- a/Sources/CLICommands/DebuggerServer.swift
+++ b/Sources/CLICommands/DebuggerServer.swift
@@ -21,6 +21,7 @@
     import SystemPackage
     import WasmKit
     import WasmKitGDBHandler
+    import WasmKitWASI
 
     struct DebuggerServer {
         var host = "127.0.0.1"
@@ -28,6 +29,7 @@
         var logLevel = Logger.Level.info
         let wasmModulePath: FilePath
         let engineConfiguration: EngineConfiguration
+        let wasiConfiguration: WASIConfiguration
 
         func run() async throws {
             let logger = {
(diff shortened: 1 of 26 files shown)


Sources/CLICommands/DebuggerServer.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if WasmDebuggingSupport
+#if WasmDebuggingSupport && !os(Windows)
```
kateinoigakukun: Why do we enable DebuggingSupport trait on Windows in the first place?

Sources/WASI/WASIBridgeToHost.swift
```diff
@@ -240,9 +240,14 @@ public final class WASIBridgeToHost: Sendable {
...
-            try fileSystemOptions.initializeStdio?(&table)
-            try fileSystemOptions.initializePreopens?(fileSystem, &table)
+        do {
+            try underlying.fdTable.withLock { table in
+                try fileSystemOptions.initializeStdio?(&table)
+                try fileSystemOptions.initializePreopens?(fileSystem, &table)
+            }
+        } catch {
+            // A throw here runs `deinit`, whose precondition requires `close()`, so release what was opened.
+            throw CleanupFailure.preserving(error, cleanup: close)
```
kateinoigakukun, line 252: Should we use it in runAndClose as well?

Sources/CLICommands/DebuggerServer.swift
MaxDesiatov: Thanks, that's cleaned up

Sources/WASI/WASIBridgeToHost.swift
MaxDesiatov: Done. `runAndClose` now goes through `withThrowing` whose failure path calls `CleanupFailure.preserving`

kateinoigakukun (approved):

PR #398, opened 2026-07-30
Replace SystemExtras with an internal, injectable platform layer in WASI

kateinoigakukun:
Part 2 of removing the swift-system / SystemExtras dependency (stacked on #397; will be retargeted to `main` once that merges). This PR deletes the `SystemExtras`/`CSystemExtras` targets and gives WASI its own platform abstraction layer. After this, swift-system remains in the manifest only for `WasmKitGDBHandler`, which a follow-up removes together with NIO.

## Internal platform layer

- New `Sources/WASI/Platform/PAL/`: a trimmed `FileDescriptor` wrapper over raw syscalls (open/read/write/pread/pwrite/seek/stat/futimens/`*at` family/readdir/clocks/poll), errno-based error handling that preserves the existing WASI errno mapping table, the vendored Windows error mapping absorbed and trimmed, and a `CWASIPlatform` C shim replacing `CSystemExtras` for wasi-libc's pointer-typed clock ids.
- **Every platform conditional is confined to `PAL/`** — no other file in the module branches on the platform. Unknown platforms compile via a stub PAL arm (host-filesystem operations fail with `ENOTSUP` at runtime) instead of hitting `#error`, so the platform-independent backends still work there.
- Guest path resolution now uses a WASI-semantics parser (`GuestPath`: `/`-separated, UTF-8) instead of host `FilePath` structural semantics, fixing a latent mismatch on Windows. Sandbox invariants (absolute-path rejection, absolute-symlink ban, symlink budget) are preserved and covered by new unit tests.

## Injectable platform seam (`@_spi(WASIPlatform)`)

For running WASI guests on platforms WasmKit doesn't know about (e.g. custom embedded systems), the abstraction boundary is exposed behind `@_spi(WASIPlatform)` so it stays out of the default API surface: `FileSystemImplementation`, `WASIEntry`/`WASIFile`/`WASIDir`/`WASIReaddirIterator`, `FdEntry`, the `WASIAbi` types they reference, `FileSystemOptions.custom(_:)`, and a `withStdio` overload taking arbitrary `WASIFile` resources (e.g. a UART-backed sink). Clocks and random generators were already injectable via non-SPI initializers. `CustomPlatformInjectionTests` runs a WASI instance entirely on injected implementations.

## Public API changes

- `FileContent.handle(CInt)`, `MemoryFileSystem.addFile(at:handle: CInt)`, and `CInt` stdio parameters (defaults 0/1/2) in `WASIBridgeToHost` initializers and `withStdio`. All descriptor-taking APIs document ownership: descriptors are borrowed and never closed by WASI.
- `WASIDir` loses the host-specific `openFile` requirement (it returned a platform descriptor and was only an implementation detail of the host backend).

## Consumers

CLICommands (except the debugger path), WasmTools, Benchmarks, and tests move to Foundation file IO. CMake drops the swift-system FetchContent. Also fixes the Windows `FILETIME`→Unix-nanoseconds conversion, which had a wrong epoch factor in SystemExtras.

## Notes for reviewers

- The sandbox path-resolution rewrite (`Open.swift`, `GuestPath`) is security-sensitive and deserves the closest look; logic is a line-for-line port with `FilePath.Component` replaced by `GuestPath.Component`.
- The unknown-platform stub arm (`PALUnsupported.swift`) cannot be compile-checked by CI (no such toolchain target); its API surface must be kept in sync with the real PAL manually.


Sources/CLICommands/Platform.swift
```diff
@@ -1,20 +1,24 @@
-import struct SystemPackage.FileDescriptor
+import Foundation
```
MaxDesiatov: I'm concerned with `CLICommands` gaining Foundation dependency. These can be Embedded Swift-compatible on macOS, Linux, and WASI. Swift System is now compatible with Embedded Swift thanks to untyped throws support. I'd rather see we drop ArgumentParser dependency in CLICommands  than add Foundation dependency.

kateinoigakukun: As long as ArgumentParser depends on Foundation, CLI cannot be Embedded compatible, IIUC. Anyway, it's easy to get rid of direct Foundation dependency here.

MaxDesiatov: That's fair, but I don't think we need 100% features of ArgumentParser either. And ArgumentParser is a separate concern from this PR, I'll look into that.

PR #405, opened 2026-08-01
Do not read a host callee as a wasm function when stepping

kateinoigakukun:
`InternalFunction` packs host and wasm functions into one word, with bit 0 set
for host functions. `var host` masks that tag off; `var wasm` does not
(`bitPattern & ~0b0` clears nothing).

`Debugger.predictNext_call` called `op.callee.assumeCompiled()` without an
`isWasm` guard, unlike its sibling `predictNext_returnCall`. The translator
emits `call` for host functions and for wasm functions in another instance
(`Translator.swift:1747`), so stepping over a call to an imported host function
read `WasmFunctionEntity.code` out of a `HostFunctionEntity` allocation.

What that produced depended on the memory it landed on:

- decoded as `.uncompiled`, tripping `assumeCompiled()`'s `preconditionFailure()`
  and aborting with `WasmKit/Function.swift:227: Fatal error`
- read past the smaller host allocation, which AddressSanitizer reports as a
  heap-buffer-overflow

Both showed up in CI as flaky `CLICommandsTests` failures on unrelated pull
requests, with signal 4, 5 or 6.

Neither operand of `call` is guaranteed compiled, and that is by design: the
engine compiles lazily in `invoke()` (`Execution.swift:684`). Only
`internalCall` carries the "already compiled" guarantee, because `compilingCall`
compiles the callee and then rewrites itself to `internalCall`.

### Fix

Add a non-trapping `InternalFunction.compiledIseq()`, and route every debugger
predictor through one `calleeEntryPc(_:)` helper that checks `isWasm`, compiles
on demand, and returns no prediction when there is nothing to step into.

This also closes a second way to reach the trap: `try? ensureCompiled(...)`
followed by `assumeCompiled()` swallowed a compilation error and then hit the
precondition. `assumeCompiled()` now remains only on the engine's `internalCall`
path, and documents the invariant it depends on.

### Verification

`DebuggerHostCallStepTests` steps over a call to an imported host function. It
crashes on every run against unpatched sources and passes with this change.

Running the full suite in a loop reproduced the crash on run 7 of 20 before the
fix, and 20 of 20 runs are clean after it.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -480,19 +480,31 @@
             return (0..<Int(op.count)).map { pc.advanced(by: Int(op.baseAddress[$0].offset)) }
         }
 
+        /// Resolves the entry PC of a call target, compiling it if needed.
+        ///
+        /// Returns `nil` when there is nothing to step into: a host function, or
+        /// a callee whose compilation fails. `call` is emitted for host
+        /// functions and for wasm functions in another instance, so neither
+        /// "is wasm" nor "is compiled" can be assumed here.
+        private mutating func calleeEntryPc(_ callee: InternalFunction) -> Pc? {
+            guard callee.isWasm else { return nil }
+            _ = try? callee.wasm.ensureCompiled(store: StoreRef(self.store))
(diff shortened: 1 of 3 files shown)


MaxDesiatov (approved):
Thanks!

PR #415, opened 2026-08-04
Add `wasmkit-cli wast` subcommand w/ new `WASTRunner` module

MaxDesiatov:
Running a `.wast` script required `swift test`, because the directive interpreter lived in the `WasmKitTests` target. It moves to `Sources/WASTRunner`, which the test suite and the new subcommand share. WAT's public `Wast*` types are renamed to `WAST*`.

In a future PR this module will be ported to Embedded Swift to run the full spec test suite on embedded platforms without requiring Swift Testing.


kateinoigakukun (approved):
Yes! This is exactly what I wanted to do next, thanks!

PR #417, opened 2026-08-06
Stop leaking and losing breakpoints while stepping

JDevlieghere:
Stepping arms a breakpoint at every address execution can reach from the current one, but nothing ever took them back down. Every step left live breakpoints behind at addresses the host never asked about, and so did every continue, which single-steps off the breakpoint it resumes from. The program then stopped where nobody had set a breakpoint.

The opposite happened too. run() resumed by calling disableBreakpoint(), which forgets the host's request as well as taking the breakpoint out of the bytecode, and runPreservingCurrentBreakpoint() only put back the address it started from, not the one its intermediate step landed on. A continue that landed on another of the host's breakpoints destroyed it, which is also why a breakpoint one instruction past the resumed-from one was never reported.

Both come from conflating two facts in one dictionary: whether a breakpoint is in the bytecode, and whether someone wants one at that address. A breakpoint traps before the instruction it replaced, so resuming has to take it out of the bytecode and cannot put it back until execution has left the slot, but the request for it outlives all of that. Keep the two apart: armedBreakpoints is what is in the bytecode, hostBreakpoints and stepBreakpoints are the two reasons to want one, and an address stays armed while either reason remains. step() now arms, runs, clears its own breakpoints and puts back the breakpoint it stepped off, which leaves runPreservingCurrentBreakpoint() as a step plus a resume.

Detach goes through a new removeAllBreakpoints() instead of walking the breakpoint map, which is no longer exposed.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -46,8 +46,21 @@
         /// Threading model of the Wasm engine configuration, cached for a potentially hot path.
         private let threadingModel: EngineConfiguration.ThreadingModel
 
-        /// Mapping from a Wasm address of a breakpoint to a corresponding iseq code slot.
-        package private(set) var breakpoints = [Int: UInt64]()
+        /// Breakpoints currently present in the bytecode, keyed by resolved Wasm address. The value
+        /// carries both what to restore and where, so taking one down never has to resolve again.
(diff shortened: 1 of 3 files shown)


kateinoigakukun (approved):
Thanks!

MaxDesiatov (approved):
LGTM 👍

kateinoigakukun:
Okay, GitHub Actions is degraded, let's merge once it's repaired

PR #418, opened 2026-08-14
Exit with the client and detach from the guest

JDevlieghere:
A connection ending without `k` or `D` sent the server back to `accept()`, where it waited for a second client indefinitely while holding the guest. The guest only advances on packets from the connection that drove it and cannot be restarted, so serve one connection and exit with it.

Treat an unannounced disconnect as a detach and implement `QSetDetachOnError` so the client can pick. LLDB sends it at launch from `target.detach-on-error`, which defaults to true, and debugserver detaches a running inferior when its packet connection drops. Detaching reuses the `D` path: drop the breakpoints, then resume.

`Debugger.run()` traps with "Restarting a Wasm module from the debugger is not implemented yet" in `.entrypointReturned`, so a detach from a guest that already returned stops after dropping the breakpoints.

Closing a connection without `k` now exits within a second instead of waiting indefinitely, and `QSetDetachOnError:0` exits without resuming. The resume is blocking, so a guest that never stops still outlives its client.

### DIFF
--- a/Sources/CLICommands/DebuggerServer.swift
+++ b/Sources/CLICommands/DebuggerServer.swift
@@ -35,29 +35,30 @@
             defer { listener.close() }
             logger.trace("Debugger server listening on port \(port)")
 
-            // A GDB stub serves one client at a time: accept connections
-            // sequentially until the client asks the target to shut down.
-            serving: while true {
-                let connection = try listener.accept()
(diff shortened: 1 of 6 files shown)


kateinoigakukun (approved):
Thanks! Merging as thr CI failure is irrelevant to this change.

PR #420, opened 2026-08-26
Add support for LLDB's updated `qWasmGlobal` packet

JDevlieghere:
Currently the qWasmGlobal takes a frame index, making it impossible to request a global for a module instance that's not active on the stack. See #llvm-project/212833 for more details.

LLDB now supports instance owning a global with an `instance:<id>` field, negotiated through `qWasmGlobalInstance+`. This PR adds support for the new variant of the packet. The legacy frame variant stays supported for (older) debuggers that do not support it.

Also derive the executable code offset from the address tag bits, so the instance id matches.

Assisted-by: Claude

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -390,8 +390,10 @@
             throw Error.stackFrameIndexOOB(frameIndex)
         }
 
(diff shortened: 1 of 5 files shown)


JDevlieghere:
CC @felipepiovezan

MaxDesiatov (approved):
This makes sense, thank you!

kateinoigakukun (approved):

PR #386, opened 2026-07-24
wasi-threads: add `thread-spawn` over host OS threads and shared memory

MaxDesiatov:
New `thread-spawn` import per wasi-threads spec spawns a POSIX thread and re-instantiates the module over the group's shared memory and runs `wasi_thread_start`. A trap or `proc_exit` on any thread signals a shared flag that the others observe at call boundaries and `memory.atomic.wait`, then unwind.

PR #423, opened 2026-09-01
Add WASI Threads thread-spawn support

kateinoigakukun:
## Summary
Rework https://github.com/swiftwasm/WasmKit/pull/386

- add the WasmKitWASIThreads product, public shared-memory backing, and detached pthread worker startup
- add CLI WASI Threads opt-in with explicit `--feature threads` proposal gating
- add unit, shared-memory, option, and subprocess lifecycle coverage
- address review feedback on resource limiting, documentation, feature configuration, and test ownership

## Validation
- `swift build --product wasmkit-cli`
- `swift test --filter RunOptionsTests`
- `swift test --filter WASIThreadsTests`
- `swift test --filter WASIThreadsSubprocessTests`


README.md
```diff
@@ -130,7 +130,7 @@ Proposals are grouped by their [phase](https://github.com/WebAssembly/meetings/b
 | Feature | Status | WasmKit version |
 |---------|--------|-----------------|
 | [WASI 0.1](https://github.com/WebAssembly/WASI/tree/wasi-0.1) | 🚧 [Majority of syscalls implemented](https://github.com/swiftwasm/WasmKit/blob/d9b56a7b3f979a72682c0d37f6cc71b3493dae65/Tests/WASITests/IntegrationTests.swift#L31) | [0.0.2] |
-| [WASI Threads](https://github.com/WebAssembly/wasi-threads) | ❌ Not implemented | |
+| [WASI Threads](https://github.com/WebAssembly/wasi-threads) | ✅ Implemented | `wasmkit-cli run --feature threads --wasi-threads` |
```
MaxDesiatov: Do you think that passing both should be required? I feel like `--feature threads` should be enough, and we can detect that it's WASIp1 threads from the list of imports?

Documentation/WASIThreads.md
```diff
@@ -0,0 +1,30 @@
...
+limit of one permits no children. IDs are allocated from 1 through
+`0x1FFFFFFF`, are never reused, and failed attempts after reservation also
+consume an ID.
+
+This is intentionally a process-oriented CLI capability, not a general
+long-lived embedding API. A worker trap or `proc_exit` terminates the host
+process; returning from `_start` also terminates it, allowing the operating
+system to clean up detached workers. It requires a 64-bit macOS or Linux host,
+the `MultiThread` package trait, direct-threaded execution, the threads
+feature, mprotect bounds checking, and an engine without an interceptor.
```
MaxDesiatov: The last sentence as written seems more relevant for package/library WasmKit users, which is not consistent with the first sentence saying it's a CLI-only capability? I'd simplify it this way:
```suggestion
system to clean up detached workers. It requires a 64-bit macOS or Linux host,
and the `MultiThread` package trait when building the CLI product.
```

Sources/CLICommands/Run.swift
```diff
@@ -44,6 +53,41 @@ package struct Run: AsyncParsableCommand {
     )
     var signpost: Bool = false
 
+    @Flag(name: .customLong("wasi-threads"), help: "Enable WASI Threads")
+    var wasiThreads = false
```
MaxDesiatov: Same as `README.md` comment below: I feel this is redundant, we can auto-detect if from imports when `--feature threads` is passed.
```suggestion
```

Sources/CLICommands/Run.swift
```diff
@@ -44,6 +53,41 @@ package struct Run: AsyncParsableCommand {
...
+            case .exceptionHandling: .exceptionHandling
+            }
+        }
+    }
+
+    @Option(
+        name: .customLong("feature"),
+        help: "Enable a WebAssembly proposal feature (memory64, reference-types, threads, tail-call, simd, exception-handling)"
+    )
+    var features: [Feature] = []
```
MaxDesiatov, line 95: Did we not have any of these enabled by default previously? It would be confusing if CLI supported EH or SIMD by default, and in the new release we'd start requiring explicit options for it.

Sources/CLICommands/Run.swift
kateinoigakukun: I think auto-enabling `--feature threads` based on `--wasi-threads` makes sense but not the opposite side, because wasi-threads depends on core threads proposal but not the other way.

Sources/CLICommands/Run.swift
kateinoigakukun: Good point. We should keep the default feature set consistent with the one in WasmParser.

Sources/CLICommands/Run.swift
```diff
@@ -456,3 +553,15 @@ extension Run {
         return (functionName, parameters)
     }
 }
+
+private func terminateProcess(_ code: Int32) -> Never {
```
MaxDesiatov: `fatalError("WASI Threads is unavailable on this platform")` in this body makes this function specific to the threads implementation, which IMO should be reflected in its name. The current name seems too general.

```suggestion
private func wasiThreadsTerminateProcess(_ code: Int32) -> Never {
```

PR #424, opened 2026-09-02
Report the guest's exit status instead of dropping the connection

JDevlieghere:
A WASI `proc_exit` reaches the target as a `WASIExitCode` thrown out of the host call. It escaped the packet handler, which tore down the connection, so LLDB saw the guest vanish rather than the status it ended with.

Catch it wherever the guest is resumed and record it as a new `exited` debugger state, reported as `W<status>`. The state persists so a `c` after the exit reports it again instead of restarting.

### DIFF
--- a/Tests/WasmKitGDBHandlerTests/TestHandler.swift
+++ b/Tests/WasmKitGDBHandlerTests/TestHandler.swift
@@ -1,0 +1,28 @@
+#if WasmDebuggingSupport
+
+    import WasmKitWASI
+
+    @testable import WasmKit
+    @testable import WasmKitGDBHandler
+
+    /// Runs `body` against a handler debugging `wasmBinary`, stopped at its entrypoint.
+    ///
+    /// Closes on both paths because WASIBridgeToHost's deinit preconditions on close() having run.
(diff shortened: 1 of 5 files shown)


Tests/WasmKitGDBHandlerTests/TestHandler.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    import WasmKitWASI
+
+    @testable import WasmKit
+    @testable import WasmKitGDBHandler
+
+    /// Runs `body` against a handler debugging `wasmBinary`, stopped at its entrypoint.
+    ///
+    /// Closes on both paths because WASIBridgeToHost's deinit preconditions on close() having run.
+    func withDebuggedHandler(debugging wasmBinary: [UInt8], _ body: (WasmKitGDBHandler) throws -> Void) throws {
```
MaxDesiatov: Could this and `withHandler` be generalised? https://github.com/swiftwasm/WasmKit/blob/1595da0f159b690bfcf5a6035d85705ed9a139d8/Tests/WasmKitGDBHandlerTests/WasmKitGDBHandlerTests.swift#L64

JDevlieghere:
I don't think the failure is related to my change, but tbh it's somewhat hard to tell from the logs.

MaxDesiatov (approved):
Thank you! 🙏

PR #425, opened 2026-09-04
Report a stop at the start of the instruction run

JDevlieghere:
The translator only mapped instructions that emitted bytecode, skipping `local.get`, `i32.const`, and `block`. Stops mapped to the next emitted instruction instead, landing mid-statement and causing steps to overshoot. Track each bytecode slot's instruction run and report its start. Make `BreakpointState.reportedPc` the single source of truth so stop replies and `qWasmCallStack` agree.

The debugger disassembles from the run start, so it can set breakpoints on non-emitting instructions within the run. Because every instruction in a run shares one bytecode slot, track requested addresses per slot so they remain distinct, and report successive breakpoints on the same slot without resuming execution. Removing a breakpoint now requires the originally requested address.

Attribute stops by resolved address rather than reported address. If a breakpoint was set before code was compiled, it may have resolved to a later instruction. When that code is later compiled and stepped onto, matching on reported address falsely reported a breakpoint hit, causing LLDB to abort its step plan.

### DIFF
--- a/Sources/WasmKit/Execution/Debugger.swift
+++ b/Sources/WasmKit/Execution/Debugger.swift
@@ -5,7 +5,10 @@
     package struct Debugger: ~Copyable {
         package struct BreakpointState {
             let iseq: Execution.Breakpoint
+            /// Wasm address the engine stopped on.
             package let wasmPc: Int
+            /// Wasm address reported for this stop.
+            package var reportedPc: Int
         }
 
         package enum State {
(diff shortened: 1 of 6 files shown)


JDevlieghere:
This PR is larger and more complex than the previous ones, but gets 15 more tests passing with WasmKit (from 562 to 577).

kateinoigakukun (approved):

JDevlieghere:
The test failure looks unrelated (again). Should the Windows dependency checkout include component-model?

MaxDesiatov:
Android jobs are historically flaky (IIRC at the emulator step?), but I haven't seen too many errors on the Windows job before, probably a rare flake there too, I restarted them all.

MaxDesiatov (approved):
Great that we're flushing out these edge cases, thank you!

PR #429, opened 2026-09-09
Add a wasmi-benchmarks replay harness under Benchmarks/

kateinoigakukun:
Vendors wasmi-labs/wasmi-benchmarks (with its res/rust submodule) and adds a WasmiBenchmarks executable that replays the execute suite and CoreMark against the WasmKit library with the same inputs and result assertions.

### DIFF
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -75,3 +75,17 @@ package.targets += [
         ]
     ),
 ]
+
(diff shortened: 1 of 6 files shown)


MaxDesiatov (approved):
`WasmiBenchmarks.swift‎` looks very ad hoc, but probably fine as a one-off.

PR #430, opened 2026-09-09
Recommend --omit-frame-pointers for Linux release builds and use it for release artifacts

kateinoigakukun:
Every `wasmkit_tc_*` handler on x86_64 begins `push %rbp; mov %rsp,%rbp` and ends `pop %rbp` before `jmp *%rax`.
Omitting the frame-pointer prologue/epilogue to reduce a few instructions per instruction handler.

Measured on my Linux box (2 runs each): CoreMark 2868 -> 3081; fibonacci-iter -27%.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,11 @@ jobs:
       - run: |
           VERSION=$(echo ${{ github.ref_name }} | sed 's/^v//')
           echo "WASMKIT_VERSION=${VERSION}" >> "$GITHUB_ENV"
-      - run: ./build-exec ./Utilities/build-release.py -o wasmkit-x86_64-swift-linux-musl.tar.gz  -s $WASMKIT_VERSION -- --swift-sdk x86_64-swift-linux-musl
+      # x86-64 only: By default, fp prologue/epilogue are emitted for VM instruction handlers even though
+      # they are "leaf" function (in LLVM, a function whose only call is a guaranteed tail call counts as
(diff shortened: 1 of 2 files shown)


MaxDesiatov (approved):

.github/workflows/release.yml
```diff
@@ -47,8 +47,8 @@ jobs:
       - run: |
           VERSION=$(echo ${{ github.ref_name }} | sed 's/^v//')
           echo "WASMKIT_VERSION=${VERSION}" >> "$GITHUB_ENV"
-      - run: ./build-exec ./Utilities/build-release.py -o wasmkit-x86_64-swift-linux-musl.tar.gz  -s $WASMKIT_VERSION -- --swift-sdk x86_64-swift-linux-musl
-      - run: ./build-exec ./Utilities/build-release.py -o wasmkit-aarch64-swift-linux-musl.tar.gz -s $WASMKIT_VERSION -- --swift-sdk aarch64-swift-linux-musl
+      - run: ./build-exec ./Utilities/build-release.py -o wasmkit-x86_64-swift-linux-musl.tar.gz  -s $WASMKIT_VERSION -- --swift-sdk x86_64-swift-linux-musl --omit-frame-pointers
+      - run: ./build-exec ./Utilities/build-release.py -o wasmkit-aarch64-swift-linux-musl.tar.gz -s $WASMKIT_VERSION -- --swift-sdk aarch64-swift-linux-musl --omit-frame-pointers
```
MaxDesiatov: If it's beneficial only on x86, why are we applying it for aarch64?

PR #433, opened 2026-09-09
Fuse integer comparisons into conditional branches

kateinoigakukun:
Every integer comparison wrote its i32 result into a stack slot and the following `br_if` immediately reloaded that slot: one extra dispatch, one store and one load per loop back-edge

Before that can be done cleanly there is a generator-hygiene problem to fix: LLVM's function-merging pass folded 40 bit-identical handler bodies (`i32.load`/`f32.load`, `i32.store`/`f32.store`, the `i64.atomic.*` width variants, pure slot moves) and left one-instruction `b <other>` thunks that the handler table pointed at, so those opcodes paid two branches per dispatch instead of one.

So this PR first makes that sharing explicit in the generator, then adds 20 fused compare-branch instructions with a one-slot immediate and rewrites a comparison immediately consumed by `br_if` or `if` into the fused form.

### Metrics

Direct threading, base and new built from the same sources except this change, run back to back; `min` ms:

|---|---|---|---|
| regex-redux | 0.141 | 0.111 | **0.787** |
| fibonacci-iter | 5.421 | 4.729 | **0.872** |
| json-parse | 41.227 | 36.646 | **0.889** |
| nbody | 91.154 | 85.271 | 0.935 |
| word-count | 5.344 | 5.067 | 0.948 |
| argon2 | 208.564 | 200.641 | 0.962 |
| counter-local | 2.512 | 2.424 | 0.965 |
| counter-global | 1.673 | 1.626 | 0.972 |
| fibonacci-rec | 15.918 | 15.470 | 0.972 |
| sort-dyn | 239.010 | 232.931 | 0.975 |
| **`execute/*` geomean** | | | **0.9788** |

+3.4% coremark score


MaxDesiatov (approved):
Nice, thanks!

PR #436, opened 2026-09-12
Make wasm-to-wasm return call-free and merge frame initialization into one copy

kateinoigakukun:
**Return.** `popFrame` chased `sp[-3]` -> function entity -> instance twice (once for the frame being popped, once for the caller), compared them and conditionally switched `md`/`ms`. That ran on every return, although caller and callee are in the same module for nearly every return, and because switching ends in a call (`wasmkit_trap_guard_set_current_memory`) the handler carried a prologue and epilogue that every return paid.

Whether the return needs that switch is known at the call, so the frame header can carry it instead. `Pc` points at 8-byte `CodeSlot`s, so bit 0 of a saved PC is always zero; the call sets it exactly when callee and caller instances differ, which is the comparison `invokeWasmFunction` already makes, and `internalCall`/`compilingCall` leave it clear by construction. `_return` now tests that one bit, and when it is clear the slot *is* the return `Pc`, so there is nothing to mask off. The cross-instance case became its own handler, `returnCrossInstance`, which the translator never emits: `_return` hands back its head slot as the next handler with `sp` and `pc` untouched, so direct threading musttails into it and the token-threaded loop dispatches to it on the next iteration.

This also fixes a bug on `main`: `popFrame` restored the memory of the frame it was popping instead of the caller's, so after a cross-instance call `md`/`ms` still pointed at the callee's linear memory. The caller's next store went into the callee's memory, and a store past the callee's size trapped although it was in bounds for the caller. `Tests/WasmKitTests/ExtraSuite/cross_instance_return.wast` covers the three shapes (plain store, high store, high store after the callee grew its own memory).

**Frame init.** Entering a function `memset` the locals and then `memcpy`d the constant pool. Neither is big (16 slots on average over the suite, and `fibonacci-rec`'s only function copies two), and at that size the two libc calls cost several times the copy itself; `memmove` plus the stub reaching it was ~20% of `fibonacci-rec`'s samples. `InstructionSequence.constants` becomes `frameInit`: one contiguous image of zeroed local slots followed by the pool, built once at translation time. `initializeFrame` lays images of up to 16 slots down with straight-line overlapping head/tail copies through `SIMD2<UInt64>` (`ldr q`/`str q` on arm64, `movups` on x86-64) and falls back to `copyMemory` above that. Locals still start zeroed: the zeros are in the image. `pushFrame` loses its `numberOfNonParameterLocalSlots` argument as a side effect, so `internalCall` reads one less field off the function entity.

The 16-byte unit is what makes this a win rather than a regression: an earlier version copying slot by slot left `sort` consistently 2-4% slower, because for 16-slot images 32 scalar loads and stores lose to `memmove`'s own vector loop even including the call.

### Metrics

Direct threading, `main` and this branch built from the same sources except this change and run interleaved round by round; best of 5 rounds, `min` ms:

|---|---|---|---|
| fibonacci-rec | 15.413 | 11.429 | **0.742** |
| fibonacci-tail | 6.381 | 5.905 | **0.925** |
| sort-dyn | 230.466 | 221.615 | 0.962 |
| reverse-complement | 0.151 | 0.149 | 0.987 |
| bulk-ops | 0.402 | 0.397 | 0.988 |
| word-count | 4.983 | 4.928 | 0.989 |
| counter-param | 3.037 | 3.004 | 0.989 |
| prime-sieve | 177.480 | 175.628 | 0.990 |
| matrix-mul | 277.779 | 276.355 | 0.995 |
| json-parse | 35.581 | 35.447 | 0.996 |
| compression | 33.966 | 33.861 | 0.997 |
| counter-global | 1.195 | 1.192 | 0.997 |
| counter-local | 2.298 | 2.293 | 0.998 |
| sort | 256.084 | 255.618 | 0.998 |
| regex-redux | 0.107 | 0.107 | 1.000 |
| tiny-keccak | 0.118 | 0.118 | 1.000 |
| fibonacci-iter | 4.660 | 4.662 | 1.000 |
| spectralnorm | 82.240 | 82.308 | 1.001 |
| mandelbrot | 114.266 | 114.507 | 1.002 |
| argon2 | 195.990 | 196.421 | 1.002 |
| nbody | 82.624 | 83.922 | 1.016 |
| **`execute/*` geomean** | | | **0.9778** |

CoreMark 3375 -> 3380 (1.001x), i.e. unchanged: many of its functions have images over 16 slots, so they still take the `memcpy` path and only save the separate `memset`.

`fibonacci-rec` -26% and `fibonacci-tail` -7% are the two cases dominated by call/return bookkeeping; `fibonacci-rec` repeats to within 0.5% across all five rounds (0.738/0.738/0.738/0.738/0.743). `nbody` +1.6% is the largest case that repeats in the other direction; its profile contains no `internalCall`, no `_return` and no `memcpy` at all, and every handler on its path is instruction-identical between the two binaries, so what moved is code placement.

Handler instruction counts, arm64 release. The full `_wasmkit_tc_*` diff against `main` is these four and nothing else:

| handler | before | after |
|---|---|---|
| `_return` | 48, with a stack frame | **10**, no frame (6 on the intra-module path) |
| `returnCrossInstance` | (new) | 41 (the cold half of the old handler) |
| `internalCall` | 88 | 126 |
| `compilingCall` | 128 | 167 |

The two call handlers grow because four image-size tiers are laid out straight-line, but the executed path is shorter: the two library calls are gone for images of 16 slots or fewer. `_return` on x86-64 is 8 instructions on the intra-module path, likewise with no frame, ending in `jmp rax`.

WasmKit module `__TEXT` 640,687 -> 642,925 bytes (+2.2 KB).

### Tests

`SpectestTests` (direct and token threading) and `WasmKitTests`, including the new `cross_instance_return.wast`.

### DIFF
--- a/Sources/WasmKit/Execution/Execution.swift
+++ b/Sources/WasmKit/Execution/Execution.swift
@@ -110,47 +110,114 @@ struct Execution: ~Copyable {
         return Backtrace(symbols: symbols)
     }
 
-    private func initializeConstSlots(
-        sp: Sp, iseq: InstructionSequence,
-        numberOfNonParameterLocalSlots: Int
-    ) {
-        // Initialize the locals with zeros (all types of value have the same representation)
-        sp.initialize(repeating: UntypedValue.default.storage, count: numberOfNonParameterLocalSlots)
-        if let constants = iseq.constants.baseAddress {
-            let count = iseq.constants.count
-            sp.advanced(by: numberOfNonParameterLocalSlots).withMemoryRebound(to: UntypedValue.self, capacity: count) {
-                $0.initialize(from: constants, count: count)
+    /// Lays the callee's frame-initialisation image (zeroed locals followed by
+    /// the constant pool) over the new frame's local/constant area.
+    ///
+    /// The image is one contiguous buffer built at translation time, so this is a
+    /// single copy rather than a `memset` of the locals plus a `memcpy` of the
+    /// pool -- and small images are copied inline. That matters more than it
+    /// looks: even though the images are tiny for some trivial functions, a libc call
+    /// costs several times the copy itself.
+    @inline(__always)
+    private func initializeFrame(sp: Sp, iseq: InstructionSequence) {
+        let image = iseq.frameInit
+        guard let src = image.baseAddress else { return }
+        let count = image.count
+        let dst = UnsafeMutableRawPointer(sp).assumingMemoryBound(to: UntypedValue.self)
+        // Straight-line, overlapping head/tail copies. Written out rather than
+        // looped so that no loop remains for the optimizer to turn back into a
+        // `memcpy` call.
+        if count >= 8 {
+            if count > 16 {
+                // `copyMemory` rather than `update(from:count:)`: the latter emits
+                // an overlap check that is dead here (the image lives in the iseq
+                // allocation, never on the VM stack).
+                UnsafeMutableRawPointer(dst).copyMemory(
+                    from: UnsafeRawPointer(src), byteCount: count * MemoryLayout<UntypedValue>.stride
+                )
+                return
             }
+            Self.copy8(dst, src, 0)
+            Self.copy8(dst, src, count - 8)
+        } else if count >= 4 {
+            Self.copy4(dst, src, 0)
+            Self.copy4(dst, src, count - 4)
+        } else if count >= 2 {
+            Self.copyPair(dst, src, 0)
+            Self.copyPair(dst, src, count - 2)
+        } else if count == 1 {
+            dst[0] = src[0]
         }
     }
 
+    /// Copies two slots as one 16-byte unit, so this lowers to a single vector
+    /// load/store pair -- `ldr q`/`str q` on arm64, `movups` on x86-64 -- rather
+    /// than two scalar ones.
+    @inline(__always)
+    private static func copyPair(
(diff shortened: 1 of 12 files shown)


MaxDesiatov:
`check-autogen-diff (pull_request)` job failure seems trivial (some whitespace diff?)

Sources/WasmKit/Execution/Execution.swift
```diff
@@ -110,47 +110,116 @@ struct Execution: ~Copyable {
...
+    /// Lays the callee's frame-initialisation image (zeroed locals followed by
+    /// the constant pool) over the new frame's local/constant area.
+    ///
+    /// The image is one contiguous buffer built at translation time, so this is a
+    /// single copy rather than a `memset` of the locals plus a `memcpy` of the
+    /// pool -- and small images are copied inline. That matters more than it
+    /// looks: the images are tiny (2 slots for `fibonacci-rec`'s only function,
+    /// ~16 on average across the wasmi-benchmarks suite), so a libc call costs
+    /// several times the copy itself. Before this was inlined, `memmove` plus the
+    /// stub that reaches it took ~20% of `fibonacci-rec`'s samples on arm64 macOS.
```
MaxDesiatov: Can this comment be cleaned up in terms of structure and wording? It's quite verbose.

PR #438, opened 2026-09-13
Relink local.tee and v128 producers, add copy peepholes and i64.eqz branch fusion

kateinoigakukun:
`local.set` already avoided a copy when its value came from the instruction just emitted: the translator rewrote that instruction to write the local's slot directly. `local.tee` and `v128` values were excluded, so every `local.tee` in a hot loop still paid a `copyStack` into the local. Neither exclusion is needed. The value a `local.tee` leaves on the stack is pushed back as `.local(index)`, the same alias of the local's slot that `local.get` produces, so it does not matter whether the slot was filled by a copy or by the producer. A write to the same local before that value is consumed is already materialized by `preserveLocalsOnStack`. A `v128` producer writes both of its slots from one base register, so redirecting that register relinks the whole value.

Relinking used to be kept safe by resetting the last emission defensively: every pop, `local.get`, `drop`, `loop` entry and `else` dropped it. Those resets also threw away legal candidates. The last emission now records the register it writes, and `relinkLastInstructionResult(from:to:)` checks the preconditions directly: the last emission produced exactly the value being set, nothing has been emitted since, and no label is pinned after it. A pop keeps the last emission unless it pops that emission's own result. The reset at `end` stays, because ending a `loop` pins no label and a producer inside the body must not be relinked from after the loop.

Two smaller changes ride along. `emitCopyStack` drops `copy A <- B` when the previous instruction is `copy B <- A` and no label is pinned between them. And `i64.eqz` followed by `br_if` or `if` now fuses into the new `brIfI64Eqz` / `brIfI64Nez`. `brIf` and `brIfNot` read only the low 32 bits of the condition slot, which cannot stand in for a 64-bit zero test (`0x1_0000_0000` has a zero low half).

With these changes the compiler inlines `ISeqBuilder.emit` into every specialization of the `pop*Emit` helpers, which grows translator text by about 100 KB. The translator's `emit` wrapper is therefore split into one outlined overload per argument shape (instruction only, with a result, with a fusable condition), which keeps the unused arguments out of each call.

`counter-param` loop body, 4 VM instructions per iteration before and 3 after:

```
before                                after
0x00: reg:4  = copy reg:-4            0x00: reg:4  = copy reg:-4
0x02: reg:4  = i32.sub reg:4, reg:0   0x02: reg:-4 = i32.sub reg:4, reg:0
0x04: reg:-4 = copy reg:4             0x04: reg:4  = copy reg:-4
0x06: reg:4  = copy reg:-4            0x06: br_if reg:-4, -6 ; 0x2
0x08: br_if reg:-4, -8 ; 0x2          0x08: reg:-4 = copy reg:4
0x0a: reg:-4 = copy reg:4             0x0a: return
0x0c: return
```

### Metrics

Base and branch built from the same sources except this change, run interleaved round by round; best of 5 rounds, `min` ms, Apple M4 Max.

|---|---|---|---|
| prime-sieve | 159.756 | 107.710 | **0.674** |
| counter-local | 2.259 | 1.643 | **0.727** |
| mandelbrot | 108.256 | 80.105 | **0.740** |
| counter-param | 2.974 | 2.230 | **0.750** |
| sort | 243.555 | 188.615 | **0.774** |
| compression | 34.133 | 26.597 | **0.779** |
| nbody | 72.266 | 60.569 | 0.838 |
| json-parse | 34.287 | 28.848 | 0.841 |
| regex-redux | 0.100 | 0.086 | 0.860 |
| sort-dyn | 199.891 | 173.603 | 0.868 |
| argon2 | 205.079 | 180.711 | 0.881 |
| word-count | 4.713 | 4.159 | 0.882 |
| tiny-keccak | 0.116 | 0.103 | 0.888 |
| matrix-mul | 254.373 | 237.170 | 0.932 |
| counter-global | 1.221 | 1.153 | 0.944 |
| reverse-complement | 0.143 | 0.136 | 0.951 |
| fibonacci-rec | 11.364 | 10.946 | 0.963 |
| bulk-ops | 0.399 | 0.397 | 0.995 |
| fibonacci-iter | 4.633 | 4.614 | 0.996 |
| spectralnorm | 75.896 | 76.224 | 1.004 |
| fibonacci-tail | 5.455 | 5.514 | 1.011 |
| **`execute/*` geomean** | | | **0.8657** |

CoreMark 3611 -> 3844 (**1.065x**). Token threading, best of 5 rounds: geomean **0.9175**, CoreMark 1877 -> 2018 (1.075x). Eager startup (parse + instantiate of the seven wasmi `startup` modules, best of 4 rounds): geomean 1.003.

`copyStack` instructions emitted for CoreMark and the Rust benchmark modules (`wasmkit-cli explore`). Total VM instructions over the same modules go from 219,235 to 197,969 (-9.7%).

| module | before | after | delta |
|---|---|---|---|
| coremark | 405 | 265 | -34.6% |
| argon2 | 2119 | 662 | -68.8% |
| compression | 1536 | 771 | -49.8% |
| json-parse | 1920 | 1067 | -44.4% |
| mandelbrot | 601 | 261 | -56.6% |
| matrix-mul | 669 | 305 | -54.4% |
| nbody | 723 | 337 | -53.4% |
| prime-sieve | 602 | 263 | -56.3% |
| regex-redux | 36376 | 22667 | -37.7% |
| reverse-complement | 1226 | 735 | -40.0% |
| sort | 894 | 454 | -49.2% |
| sort-dyn | 1086 | 544 | -49.9% |
| spectralnorm | 677 | 317 | -53.2% |
| tiny-keccak | 706 | 324 | -54.1% |
| word-count | 910 | 329 | -63.8% |
| **total** | **50450** | **29301** | **-41.9%** |

`brIfI64Eqz` and `brIfI64Nez` are 11 instructions each on arm64, the same shape as `brIf` with `ldr x9` in place of `ldr w9`.

| | before | after | delta |
|---|---|---|---|
| WasmKit `__TEXT` (arm64) | 630,309 | 642,413 | +12,104 |
| WasmKit `__DATA` (arm64) | 54,761 | 55,289 | +528 |
| `wasmkit-cli` `__text` (arm64) | 2,032,672 | 2,042,972 | +10,300 |


kateinoigakukun:
Split into #439 (local.tee / v128 relinking), #440 (i64.eqz branch fusion), #441 (inverse-copy peephole) and #442 (explicit relink preconditions), each measured on its own.

PR #440, opened 2026-09-14
Fuse i64.eqz into br_if and if

kateinoigakukun:
`i32.eqz` followed by `br_if` or `if` is already folded into a plain `brIf` / `brIfNot` on the input, with the polarity flipped. `i64.eqz` was not, because `brIf` and `brIfNot` read only the low 32 bits of the condition slot (`sp[i32:]`), and a 64-bit zero test cannot be answered from those: `0x1_0000_0000` has a zero low half and is not zero. This adds `brIfI64Eqz` and `brIfI64Nez`, which read the whole slot, and records `i64.eqz` as a fusion candidate the same way `i32.eqz` is, so `i64.eqz` + `br_if` becomes `brIfI64Eqz` and `i64.eqz` + `if` becomes `brIfI64Nez`.

Both handlers are 11 instructions on arm64, the same shape as `brIf` with `ldr x9` in place of `ldr w9`.

`fibonacci-iter`, where the fused test guards entry to the loop:

```
before                                     after
0x06: reg:2 = copy reg:-4                  0x06: reg:2 = copy reg:-4
0x08: reg:7 = i64.eqz reg:2                0x08: br_if.i64.eqz reg:2, +10 ; 0x14
0x0a: br_if reg:7, +10 ; 0x16              0x0a: reg:7 = i64.add reg:0, reg:1
0x0c: reg:7 = i64.add reg:0, reg:1         ...
```

Stacked on #439.

### Metrics

Base and branch built from the same sources except this change, run interleaved; minimum over 10 fresh-instance subrounds from 2 alternating process pairs, `min` ms, Apple M4 Max.

| case | before | after | ratio |
|---|---|---|---|
| fibonacci-tail | 6.075 | 5.589 | 0.920 |
| bulk-ops | 0.439 | 0.410 | 0.934 |
| tiny-keccak | 0.110 | 0.104 | 0.945 |
| json-parse | 29.940 | 28.623 | 0.956 |
| fibonacci-rec | 11.411 | 10.998 | 0.964 |
| prime-sieve | 109.092 | 106.575 | 0.977 |
| counter-global | 1.278 | 1.263 | 0.988 |
| counter-local | 1.816 | 1.801 | 0.992 |
| compression | 26.956 | 26.760 | 0.993 |
| word-count | 4.300 | 4.281 | 0.996 |
| counter-param | 2.443 | 2.436 | 0.997 |
| sort | 189.606 | 189.598 | 1.000 |
[...]

PR #441, opened 2026-09-14
Drop a copy that writes back the value the previous copy just read

kateinoigakukun:
`emitCopyStack` already dropped `copy A <- A`. It now also drops `copy A <- B` when the instruction just emitted is `copy B <- A`: the second copy writes back the value the first one just read, so `A` already holds it. The previous copy is only considered while it is still the last thing in the buffer and no label is pinned at the insertion point, so the second copy cannot be reached without the first having run. `emitWithLabel` appends a branch without recording a last emission, which keeps a copy before a branch from being paired with one in its landing pad.

CoreMark, the tail of function 1:

```
before                                     after
0x7e: reg:14 = copy reg:-5                 0x7e: reg:14 = copy reg:-5
0x80: reg:-5 = copy reg:14                 0x80: return
0x82: return
```

Stacked on #440.

### Metrics

Base and branch built from the same sources except this change, run interleaved; minimum over 10 fresh-instance subrounds from 2 alternating process pairs, `min` ms, Apple M4 Max.

|---|---|---|---|
| fibonacci-tail | 5.953 | 5.771 | 0.969 |
| fibonacci-iter | 4.813 | 4.713 | 0.979 |
| prime-sieve | 109.344 | 108.347 | 0.991 |
| counter-param | 2.445 | 2.431 | 0.994 |
| fibonacci-rec | 11.018 | 10.971 | 0.996 |
| word-count | 4.275 | 4.261 | 0.997 |
| matrix-mul | 238.602 | 237.956 | 0.997 |
| mandelbrot | 83.293 | 83.250 | 0.999 |
| tiny-keccak | 0.105 | 0.105 | 1.000 |
| counter-global | 1.263 | 1.264 | 1.001 |
| bulk-ops | 0.439 | 0.440 | 1.002 |
| nbody | 61.227 | 61.454 | 1.004 |
| counter-local | 1.804 | 1.812 | 1.004 |
| spectralnorm | 76.842 | 77.521 | 1.009 |
| sort-dyn | 175.110 | 176.973 | 1.011 |
| sort | 189.998 | 192.120 | 1.011 |
| reverse-complement | 0.144 | 0.148 | 1.028 |
| compression | 26.698 | 27.669 | 1.036 |
| json-parse | 28.804 | 30.119 | 1.046 |
| argon2 | 184.085 | 194.355 | 1.056 |
| regex-redux | 0.088 | 0.093 | 1.057 |
| **`execute/*` geomean** | | | **1.0087** |

CoreMark 3746 -> 3714 (0.991x).

`copyStack` instructions emitted for CoreMark and the Rust benchmark modules go from 29,383 to 29,301.

| | before | after | delta |
|---|---|---|---|
| WasmKit `__TEXT` (arm64) | 630,929 | 635,381 | +4,452 |
| `wasmkit-cli` `__text` (arm64) | 2,033,116 | 2,037,468 | +4,352 |


kateinoigakukun:
Closing: the peephole itself has no measurable gain (82 copies across the benchmark modules). The `lastCopy` bookkeeping will come with the later change that actually needs it.

PR #442, opened 2026-09-14
Check relink preconditions explicitly instead of resetting defensively

kateinoigakukun:
Relinking a producer into a local's slot rewrites the last emitted instruction, so it is only sound when that instruction really produced the value being set and nothing has been emitted or pinned since. Compare+branch fusion already checks this for itself (`fuseCompareIntoBranch` compares the candidate's result register and calls `canRewind`), but relinking relied on it indirectly: every pop, `local.get`, `drop`, constant push, `loop` entry and `else` reset the last emission, and `relinkLastInstructionResult` trusted whatever was left.

The last emission now records the register it writes, and `relinkLastInstructionResult(from:to:)` checks the three conditions itself: the last emission produced exactly the value being set, nothing has been emitted since, and no label is pinned after it. With the check in place most of the defensive resets go away. A pop resets only when it pops the last emission's own result, and `local.get`, constant pushes, `drop`, `loop` entry and `else` no longer reset at all; each of those sites now has a comment saying why none is needed. The reset at `end` stays, because ending a `loop` pins no label and a producer inside the body must not be relinked from after the loop.

With the extra field the compiler inlines `ISeqBuilder.emit` into every specialization of the `pop*Emit` helpers, which grows translator text by about 100 KB. The translator's `emit` wrapper is therefore split into one outlined overload per argument shape (instruction only, with a result, with a fusable condition).

For the execute benchmark modules (CoreMark and the Rust cases) the translated code is identical before and after. On the startup modules the relaxed resets let a few more relinks through: `ffmpeg` has 1 fewer copy and `spidermonkey` 2 fewer.

Stacked on #441.

### Metrics

Eager startup (parse + instantiate of the seven wasmi `startup` modules), interleaved, best of 4 rounds:

| module | before | after | ratio |
|---|---|---|---|
| bz2 | 10.833 | 10.727 | 0.990 |
| pulldown-cmark | 24.239 | 24.083 | 0.994 |
| spidermonkey | 541.711 | 539.622 | 0.996 |
| ffmpeg | 2160.657 | 2186.144 | 1.012 |
| coremark | 1.020 | 1.022 | 1.002 |
| argon2 | 4.269 | 4.257 | 0.997 |
| erc20 | 0.881 | 0.877 | 0.995 |
| **geomean** | | | **0.998** |

| | before | after | delta |
|---|---|---|---|
| WasmKit `__TEXT` (arm64) | 635,381 | 642,413 | +7,032 |
| WasmKit `__TEXT` (arm64), without the outlined `emit` overloads | 635,381 | 743,433 | +108,052 |
| `wasmkit-cli` `__text` (arm64) | 2,037,468 | 2,042,972 | +5,504 |


kateinoigakukun:
Closing: on top of #439 alone this changes no translated code on any benchmark module, so it has no gain of its own. The explicit relink preconditions will come with the later change that needs them.

PR #443, opened 2026-09-14
Give call_indirect a call-free fast path over a flat table buffer

kateinoigakukun:
Stacked on #439.

Every `call_indirect` made two out-of-line calls: `prepareForIndirectCall` (instance chase to find the table, decode a `[Reference]` enum, type check), then `invoke`, which itself called `ensureCompiled` unconditionally before pushing the frame. The handler had a stack frame to make them.

This change splits `call_indirect` into a fast handler with no call in it and a `callIndirectSlow` handler for everything else.

- **Tables store raw words.** `TableEntity` keeps its elements in a flat `UnsafeMutablePointer<UInt64>` in the `UntypedValue` encoding (bit 63 is null, otherwise the entity address, which for a funcref is the `InternalFunction` bit pattern). Resolving an entry is one load, and a single signed compare rejects both null and out of bounds. `Reference` is rebuilt only at `table.get`, the public `Table` API and element-segment init.
- **The table and the caller's instance are baked into the immediate.** Both are fixed for the lifetime of the translated function, so neither handler chases `sp[-3] -> function -> instance -> tables[i]`.
- **The fast handler hands off instead of handling rare cases.** It resolves the entry, checks the host-function tag, the interned type, the callee's code state and that the callee is in the same instance, then lays the frame out inline (`pushFrameWithoutCalling`). Any failed check returns `sp` unchanged and `pc` rewound to the immediate, dispatching to `callIndirectSlow` through `Engine.callIndirectSlowSlot`, the same way `_return` reaches `returnCrossInstance`. The slow handler resolves again from scratch, so trap order (out of bounds, null, type mismatch) lives in one place. It also covers host callees, uncompiled callees, cross-instance callees, VM stack overflow and frame-init images too large to copy without `memcpy`.

`return_call_indirect` gets the same inline resolution but still enters through `tailInvoke`.

`callIndirect` on arm64:

| | before | after |
|---|---|---|
| handler instructions | 46 | 98 |
| calls in the handler | 2 | 0 |
| stack frame | yes | no |

`callIndirectSlow` is 225 instructions and `returnCallIndirect` goes from 41 to 63. Every other handler's instruction count is unchanged.

### Metrics

Base and branch built from the same sources except this change, run interleaved; minimum over 10 fresh-instance subrounds from 2 alternating process pairs, `min` ms, Apple M4 Max.

|---|---|---|---|
| sort-dyn | 173.742 | 151.570 | 0.872 |
| mandelbrot | 83.515 | 81.252 | 0.973 |
| fibonacci-tail | 6.185 | 6.076 | 0.982 |
| json-parse | 29.872 | 29.423 | 0.985 |
| tiny-keccak | 0.103 | 0.102 | 0.990 |
| compression | 27.645 | 27.411 | 0.992 |
| spectralnorm | 76.676 | 76.287 | 0.995 |
| prime-sieve | 108.778 | 108.674 | 0.999 |
| reverse-complement | 0.136 | 0.136 | 1.000 |
| sort | 188.666 | 188.864 | 1.001 |
| bulk-ops | 0.397 | 0.398 | 1.003 |
| nbody | 61.108 | 61.330 | 1.004 |
| counter-local | 1.642 | 1.648 | 1.004 |
| matrix-mul | 237.443 | 238.428 | 1.004 |
| word-count | 4.241 | 4.259 | 1.004 |
| regex-redux | 0.085 | 0.086 | 1.012 |
| counter-param | 2.204 | 2.232 | 1.013 |
| counter-global | 1.239 | 1.264 | 1.020 |
| fibonacci-rec | 11.133 | 11.419 | 1.026 |
| fibonacci-iter | 4.343 | 4.625 | 1.065 |
| **`execute/*` geomean** | | | **0.9999** |

CoreMark 3806 -> 3775 (0.992x).

Token threading:

|---|---|---|---|
| sort-dyn | 417.206 | 372.585 | 0.893 |
| fibonacci-rec | 26.922 | 25.519 | 0.948 |
| fibonacci-iter | 12.427 | 11.812 | 0.951 |
| counter-global | 4.020 | 3.886 | 0.967 |
| word-count | 9.996 | 9.877 | 0.988 |
| mandelbrot | 123.568 | 122.735 | 0.993 |
| regex-redux | 0.265 | 0.265 | 1.000 |
| counter-local | 5.043 | 5.046 | 1.001 |
| json-parse | 68.694 | 69.286 | 1.009 |
| counter-param | 5.255 | 5.323 | 1.013 |
| prime-sieve | 265.855 | 270.211 | 1.016 |
| tiny-keccak | 0.181 | 0.184 | 1.017 |
| bulk-ops | 0.426 | 0.434 | 1.019 |
| compression | 59.621 | 61.288 | 1.028 |
| spectralnorm | 169.148 | 177.539 | 1.050 |
| sort | 450.185 | 478.083 | 1.062 |
| matrix-mul | 634.656 | 686.293 | 1.081 |
| nbody | 124.615 | 135.630 | 1.088 |
| fibonacci-tail | 10.704 | 11.761 | 1.099 |
| reverse-complement | 0.358 | 0.428 | 1.196 |
| **`execute/*` geomean** | | | **1.0228** |

CoreMark 2015 -> 1961 (0.973x).

| | before | after | delta |
|---|---|---|---|
| WasmKit `__TEXT` (arm64) | 630,167 | 634,565 | +4,398 |
| `wasmkit-cli` `__text` (arm64) | 2,032,568 | 2,035,364 | +2,796 |


kateinoigakukun:
Closing: the only gain is sort-dyn (0.872 direct) and macro plugin expansion (about 1.5% faster), while token threading regresses (geomean 1.023, CoreMark 0.973x) for a new opcode and 4.4 KB of text.

PR #469, opened 2026-09-17
Package.swift: make warnings-as-errors opt-in via `WASMKIT_WARNINGS_AS_ERRORS`

shopifyski:
### Problem

Since #282 `Package.swift` applies `.treatAllWarnings(as: .error)` unconditionally. Xcode and the `swiftbuild` build system (the default since Swift 6.4) compile package **dependencies** with `-suppress-warnings`, and the Swift driver rejects that in combination with `-warnings-as-errors`:

```
error: conflicting options '-warnings-as-errors' and '-suppress-warnings'
error: SwiftDriver WasmTypes normal arm64 com.apple.xcode.tools.swift.compiler failed with a nonzero exit code.
```

So a package or Xcode project that depends on WasmKit 0.3.x does not build. SE-0443 says warning-control settings are ignored for remote dependencies, but the new build system does not strip them yet (swiftlang/swift-package-manager#9517, swiftlang/swift-package-manager#10192, FB21347321). Until that lands in a toolchain, the only workaround for dependents is to vendor WasmKit with a patched manifest.

### Change

- `Package.swift`: apply the setting only when `WASMKIT_WARNINGS_AS_ERRORS` is set in the environment (`Context.environment`), so dependents get plain settings by default.
- `.github/workflows/main.yml`: set `WASMKIT_WARNINGS_AS_ERRORS: "1"` at the workflow level, so WasmKit's own builds keep enforcing zero warnings, which was the goal of #282.

### Verification

- `swift package dump-package` shows no `treatAllWarnings` setting by default and the full setting with the variable set.
- A minimal executable package depending on WasmKit by path, built with Swift 6.3.2 (Xcode 26) and `swift build --build-system swiftbuild`:
  - against `main`: fails with the driver error above
  - against this branch: builds and runs

Draft while CI confirms the workflow-level env reaches every job, including the `xcodebuild` matrix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kateinoigakukun (approved):
Thanks :bow:

PR #473, opened 2026-09-18
Fix Embedded Swift builds on ARM and on host platforms

kateinoigakukun:
Three independent problems, all found while bringing WasmKit up on a bare-metal ARM Cortex-M7 (a Panic Playdate, `armv7em-none-none-eabi`).

### Direct threading is selected on ARM, where it cannot work

`ThreadingModel.defaultForCurrentPlatform` only special-cased WASI. `WASMKIT_USE_DIRECT_THREADED_CODE` follows `__has_feature(swiftasynccc)`, which ARM has and RISC-V does not, so an Embedded ARM target selected `.direct` and hit the `fatalError` in `DispatchInstruction`. `Engine()` traps before parsing anything:

    fatalError("Direct threading is not supported on this platform")

Existing embedded CI targets RISC-V only, where the flag happens to be off, so this was never exercised.

### Host-only code is compiled into Embedded builds that target a host OS

`MprotectLinearMemory.swift`, `SharedMemoryStorage.swift`, `SharedMemory.swift`, `StoreAllocator.swift` and `Instances.swift` guard on the operating system alone, while the type they depend on (`Platform/SystemVirtualMemory.swift`) is guarded `#if !$Embedded`. An Embedded build targeting macOS or Linux therefore compiles the consumers while their provider is excluded:

    error: cannot find type 'SystemVirtualMemory' in scope

`SignpostTracer.swift` has the same shape via `canImport(os.signpost)`. These guards now also exclude Embedded, which is a no-op for every non-Embedded configuration.

This matters for more than tidiness: building the same code as Embedded Swift for the host is a fast way to iterate before flashing, and it is how a simulator target for an embedded product is normally built.

### `_CWasmKit.h` includes `<stdatomic.h>` before `<stdint.h>`

newlib declares the atomic typedefs in terms of `int_least8_t` and friends without including `<stdint.h>` itself, so the header fails to parse under clang with a bare-metal sysroot:

    error: unknown type name 'int_least8_t'; did you mean '__int_least8_t'?

Include order is what matters here, not include path: clang's own `stdatomic.h` defers to the platform's whenever `__STDC_HOSTED__` is 1.

### Testing

- `swift build` — clean.
- `EMBEDDED_TARGET=riscv32-none-none-eabi ./Utilities/build-embedded.sh` — unchanged from `main`.
- `EMBEDDED_TARGET=armv7em-none-none-eabi ./Utilities/build-embedded.sh` — now reaches the same point as RISC-V.
- Embedded-for-macOS-host, via a downstream project that builds its simulator target that way.

Both embedded targets still stop at `WasmKitGDBHandler` with `cannot use key path in embedded Swift` on the 6.4 release toolchain. That is pre-existing and reproduces identically on unmodified `main`; CI uses a nightly, where it does not occur.

### Provenance

Found while building a WASM-4 fantasy console runtime for the Playdate, where WasmKit runs the game carts. With these three changes the interpreter runs unmodified third-party WebAssembly on the device.


MaxDesiatov (approved):
How viable would it be to exercise this configuration on CI?

kateinoigakukun:
The Playdate's sysroot is very specific to its stack, but maybe I'll add if its build is easy to break.

